Issue #42, opened 2023-05-22
Score set edit mode doesn't include all fields


afrubin:
Many of the fields like references and external identifiers are not exposed when editing a score set that has been published.

The only elements that should be frozen after publication are the score and count data, and the contents of the `WildTypeSequence` model.

PR #44, opened 2023-05-24
MaveDB-api #42 download mapped variants button

EstelleDa:


bencap (approved):
UI changes look good to me.

PR #48, opened 2023-06-07
Enable Experiment/Scoreset Searches via Associated Publication Identifiers

bencap:
UI changes for VariantEffect/mavedb-api#67


jstone-uw (changes requested):
This is a minor thing, but would you move DefaultScoreSetTable.vue out of the layout folder (which is just for things like page layouts and toolbars) so that it's just in components? I'd also rename it to ScoreSetTable.vue and change the component name accordingly in the JavaScript code.

PR #51, opened 2023-08-07
New taxonomy

EstelleDa:
Solved https://github.com/VariantEffect/mavedb-api/issues/10 problem.


src/components/screens/ExperimentView.vue
```diff
@@ -110,14 +109,14 @@
...
 
+              <div v-if="targetGene.targetSequence.taxonomy?.url"><strong>Taxonomy ID:</strong> 
+                &nbsp;<a :href="`${targetGene.targetSequence.taxonomy.url}`" target="blank">{{targetGene.targetSequence.taxonomy.taxId}}</a> 
+                </div>
               <div v-if="targetGene.targetSequence?.sequence" style="word-break: break-word">
-                <div v-if="targetGene.targetSequence.reference?.organismName"><strong>Organism:</strong>
-                  {{ targetGene.targetSequence.reference.organismName }}</div>
-                <div v-if="targetGene.targetSequence.reference?.shortName"><strong>Reference genome:</strong>
-                  {{ targetGene.targetSequence.reference.shortName }}</div>
-                <div v-if="targetGene.targetSequence.reference?.id"><strong>Genome ID:</strong>
```
ashsny, line 118: is there anything that should replace this Genome ID/Target ID display in the new version?

src/components/screens/SearchView.vue
```diff
@@ -7,8 +7,9 @@
       <div class="mavedb-search-form">
         <span class="p-input-icon-left">
           <i class="pi pi-search" />
-          <InputText v-model="searchText" ref="searchTextInput" type="search" class="p-inputtext-sm" placeholder="Search" />
-          <Button v-if="searchText && searchText.length > 0" class="mavedb-search-clear-button p-button-plain p-button-text" @click="clear"><i class="pi pi-times"/></Button>
+          <InputText v-model="searchText" ref="searchTextInput" type="search" class="p-inputtext-sm" placeholder="Search" /> &nbsp;
+          <!--<Button v-if="searchText && searchText.length > 0" class="mavedb-search-clear-button p-button-plain p-button-text" @click="clear"><i class="pi pi-times"/></Button>-->
```
ashsny, line 11: this change seems unrelated and should probably be pulled out to a separate frontend change if you want to make it.

src/lib/item-types.js
```diff
@@ -128,6 +124,20 @@ const itemTypes = {
       }
     }
   },
+  'taxonomy': {
```
ashsny, line 127: are both of these necessary? are you going to resolve this TODO before submission?

src/lib/identifiers.js
```diff
@@ -81,4 +83,31 @@ if (s) {
   return RAW_READ_REGEX.test(s)
 }
 return false
+}
+
+export async function validateTaxonomy(s) {
```
ashsny, line 88: not sure this file is the correct place for this. this file otherwise contains only constants and functions that use them so I'm not sure making API calls in here is a good idea. is this function used anywhere other than scoreseteditor? can the code live there?

src/components/screens/ScoreSetEditor.vue
```diff
@@ -827,8 +839,9 @@ export default {
           }
         }
 
-        const referenceGenomeId = _.get(this.targetGene, 'targetSequence.reference.genome.id')
-        this.referenceGenome = this.referenceGenomes.find((rg) => rg.id == referenceGenomeId)
+        // const taxonomyId = _.get(this.targetGene, 'targetSequence.taxonomy.taxId')
```
ashsny, line 842: remove commented out code

src/components/screens/ScoreSetEditor.vue
```diff
@@ -1131,6 +1146,36 @@ export default {
       }
     },
 
+    acceptNewTaxonomy: async function() {
+      const input = this.$refs.taxonomyInput
+      console.log("111")
```
ashsny, line 1151: remove debug logging and commented out code

src/components/screens/ExperimentView.vue
EstelleDa: Reference_genome won't exist anymore. They're replaced by taxonomy ID. 
BUT, Target ID is wrong deleted! I'll add it back.

src/components/screens/SearchView.vue
EstelleDa: Sure! I think it'll be better and clearer for everyone. I'll create another pull request for this feature and change this back in this branch.

src/lib/item-types.js
EstelleDa: These two are used in `ScoreSetEditor` page. `taxonomy` is used in `taxonomies` and `taxonomy-search` is used to get the whole taxonomies and list them in dropdown menu. 
No, I'm not going to solve the TODO before submission. It's from Jeremy's initial codes in this page. I'm not sure what he prefers to modify this part in the future so just mark them as others.

src/lib/identifiers.js
EstelleDa: I didn't notice that only ScoreSetEditor needs this function. I referred to other functions such as `validateDoi` but they're used in both ExperimentEditor and ScoreSetEditor pages. I'll move it to ScoreSetEditor page.

src/components/screens/ScoreSetEditor.vue
```diff
@@ -299,23 +298,29 @@
...
-                              <div v-else class="mave-reference-genome-none">&nbsp;</div>
-                            </template>
-                            <template #option="slotProps">
-                              <div class="mave-reference-genome-name">{{ slotProps.option.shortName }}</div>
-                              <div class="mave-reference-genome-organism-name">{{ slotProps.option.organismName }}</div>
-                            </template>
-                          </Dropdown>
-                          <label :for="$scopedId('input-targetGeneReferenceGenome')">Reference genome</label>
-                        </span>
+                        If taxonomy does not appear on the option list, please enter a taxonomy ID, organism name or common name from
```
ashsny, line 301: is this text still necessary? it looks from the backend code like autocomplete searches both the local database and the ncbi database and provide both as options without the user hitting enter

EstelleDa: No, it's better to delete it.

src/components/screens/ScoreSetEditor.vue
```diff
@@ -891,6 +883,8 @@ export default {
     suggestionsForAutocomplete: function (suggestions) {
       // The PrimeVue AutoComplete doesn't seem to like it if we set the suggestion list to [].
       // This causes the drop-down to stop appearing when we later populate the list.
+
+      // list [] doesn't affect dropdown menu. API side search function should return something so that dropdown menu can get data.
```
ashsny, line 887: is this a TODO? I can't tell

src/components/screens/ScoreSetEditor.vue
```diff
@@ -153,55 +152,40 @@
...
-                        @add="acceptNewDoiIdentifier"
-                        @keyup.escape="clearDoiIdentifierSearch"
-                      >
-                        <template #chip="slotProps">
-                          <div>
-                              <span>{{ slotProps.value.identifier }}</span>
-                          </div>
-                      </template>
-                    </Chips>
+                    <AutoComplete ref="doiIdentifiersInput" v-model="doiIdentifiers"
```
ashsny, line 155: it looks like when you resolved conflicts, it reverted a bunch of code that was previously changed. you may need to do it again

src/lib/identifiers.js
```diff
@@ -1,3 +1,5 @@
+import axios from 'axios'
```
ashsny, line 1: looks like you don't need this any more?

ashsny (commented):
you should not check in the .DS_Store files - they are mac os x specific thumbnail files. I would remove them from your change and add a line to the .gitignore so they are not added in the future

src/components/screens/ScoreSetEditor.vue
EstelleDa: It seems long ago notes. I delete this one.

src/components/screens/ScoreSetEditor.vue
EstelleDa: Yes, it reverted the codes back. TAT

ashsny:
looks good, you still need to drop the .DS_Store files from your change, though

PR #68, opened 2023-11-14
Add ProgressSpinner to search page and hide table contents while xhr is loading.

### LINKED ISSUE #67: Search shows 'no results' while waiting for XHR to return.
We should instead show a spinner to indicate that fetching results is in progress. In cases where the search takes a particularly long time, this can be especially confusing to users.

ashsny:
Before this PR, the previous results are shown while a new search is loading. This also has the (intended) effect of hiding those results when a new search is initiated.

Demo: https://recordit.co/9vjBQGUAAu

Fixes #67.

### DIFF
--- a/src/components/ScoreSetTable.vue
+++ b/src/components/ScoreSetTable.vue
@@ -5,6 +5,7 @@
       :options="tableOptions"
       :scrollX="scrollX"
       :scrollY="scrollY"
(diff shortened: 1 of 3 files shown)


jstone-uw:
Looks good to me!

EstelleDa (approved):

PR #77, opened 2023-11-28
Restore mavedb documentation source files and update documentation to Sphinx 7.2.6.

ashsny:
No actual updates to the text of the documentation have yet been made. That will follow soon so that those changes made are isolated and more obvious.

This first PR simply restores the source files from https://github.com/VariantEffect/MaveDB/tree/main/docs/source and re-runs with a new version of Sphinx (7.2.6). This results in very few and mostly non-visible changes to the documentation.

Instructions to update the documentation have also been added.


EstelleDa (approved):
It looks good to me.

Issue #82, opened 2023-12-07
Mostly extraneous Node polyfills


ashsny:
Currently, the site doesn't work in Vite without Node polyfills from `vite-plugin-node-polyfills`. These shouldn't really be necessary in the browser. These are currently invoked because `csv-parse` (via `csv`) uses `Buffer`. `csv-parse` is only used in `scores.js`: https://github.com/VariantEffect/mavedb-ui/blob/1df0367e37370bc643768bfc143a7ee3dd6ef9fa/src/lib/scores.js#L2 and there are alternative libraries like `papaparse` that have no such deps. We already removed `base64url` for `universal-base64url` for the same reason in https://github.com/VariantEffect/mavedb-ui/commit/e82a2900050254327aa33af29ee2da7767875f51 following the change here: https://github.com/bbi-lab/bbi-lims-ui/commit/3717e4a201801390f61d2c394df192bd2da796d1 . Once `csv-parse` is removed, we should be able to remove the polyfills as well (although there may be other uses, tbd).

PR #83, opened 2023-12-07
Upgrade Node to 20 and swap from webpack/vue-cli to Vite.

ashsny:
This involved: 
 - explicitly adding the implicit dependency on the `uuid` package,
 - resolving the conflict `@urbn/vuex-helpers` and `vuex` by removing the former, the conflict not having shown itself until the Node version upgrade,
 - moving `index.html` and making the necessary changes to it,
 - adding a basic `vite.config.js`,
 - removing `babel` and support for old browsers (probably not strictly necessary),
 - removing `core-js` and using native js `Array.reverse` instead,
 - upgrade `eslint` and `eslint-plugin-vue` versions (unclear if necessary),
 - adding `@vitejs/plugin-basic-ssl` to generate a self-signed SSL cert,
 - adding Node polyfills to support use of `Buffer` via `csv-parse` (though see #82),
 - swap `base64url` for `universal-base64url` in service of eventually removing the above,
 - remove vestigial `d3_rebind` code that did not co-operate with Vite/Rollup,
 - changing imports in `FlexDataTable.vue` following https://github.com/bbi-lab/bbi-lims-ui/commit/205a5dc356fd475c9b52071e2e4a9ad77ace9e21 to support Rollup, and
 - `package.json` `scripts` and `README.md` updates for new dev workflows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mavedb-ui",
-  "version": "2023.2.0",
+  "version": "2023.5.0",
   "private": true,
   "scripts": {
     "build": "vite build",


jstone-uw (approved):
All these changes look exactly right. Soon (if you want) we should be able to import components like FlexDataTable from a separate npm package, which is present here in an old version; sorry you had to do some extra work to get it working with Node 20 and Vite, but the changes look good.

PR #84, opened 2023-12-11
Remove Node polyfills and fix heatmap by removing last corejs dependency.

### LINKED ISSUE #82: Mostly extraneous Node polyfills

ashsny:
Fixes #82.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,15 @@
 {
   "name": "mavedb-ui",
-  "version": "2023.2.0",
+  "version": "2023.5.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "mavedb-ui",
-      "version": "2023.2.0",
+      "version": "2023.5.0",
       "dependencies": {
         "@fontsource/raleway": "^4.5.5",
         "axios": "^0.21.1",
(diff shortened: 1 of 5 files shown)


jstone-uw:
Looks good to me -- and it's nice to get rid of lodash calls where built-in collection methods work.

jstone-uw (approved):

PR #85, opened 2023-12-12
add SendToGalaxy option

PlushZ:
Hi, I created this pull request to add an option Send to Galaxy. If MaveDB is opened from Galaxy UI, this option becomes available.

Here is a description of the process of importing data from MaveDB into Galaxy:

1. Galaxy performs a HTTP GET request to the MaveDB url and passes the parameter GALAXY_URL in this request. The value of this parameter contains the url where Galaxy will expect the response to be sent at some later time. MaveDB keeps track of this URL as long as the user navigates MaveDB website.
2. As the user navigates MaveDB, it behaves exactly as if it would if the request would have not originated from Galaxy.
3. At the point where the parameter submission would return data, MaveDB instead posts these parameters to the url that was sent in the GALAXY_URL parameter. 
4. When Galaxy receives the parameters it will run a URL fetching process in the background that will resubmit the parameters to MaveDB, and it will deposit the returned data in the Galaxy user's account.

### DIFF
--- a/src/components/screens/ScoreSetView.vue
+++ b/src/components/screens/ScoreSetView.vue
@@ -53,21 +53,20 @@
             <EntityLink entityType="scoreSet" :urn="urn" />
           </template>
         </div>
-        <div>Download files <Button class="p-button-outlined p-button-sm"
-            @click="downloadFile('scores')">Scores</Button>&nbsp;
+        <div>Download files <Button class="p-button-outlined p-button-sm" @click="downloadFile('scores')">Scores</Button>&nbsp;
           <template v-if="countColumns.length != 0">
             <Button class="p-button-outlined p-button-sm" @click="downloadFile('counts')">Counts</Button>&nbsp;
           </template>
           <template v-if="isMetaDataEmpty != true">
             <Button class="p-button-outlined p-button-sm" @click="downloadMetadata">Metadata</Button>&nbsp;
           </template>
-          <Button v-if="sendToGalaxy == '1'" class="p-button-outlined p-button-sm" @click="sendGalaxy('scores')">Send
-          Scores to
-          Galaxy</Button>&nbsp;
           <Button class="p-button-outlined p-button-sm" @click="downloadMappedVariants()">Mapped Variants</Button>&nbsp;
-          <Button v-if="sendToGalaxy == '1'" class="p-button-outlined p-button-sm" @click="sendGalaxy('mappedVariants')">Send
-          Mapped Variants to
-          Galaxy</Button>&nbsp;
+        </div>
+        <div v-if="requestFromGalaxy == '1'"><br>Send files to <a :href="`${this.galaxyUrl}`">Galaxy</a> <Button class="p-button-outlined p-button-sm" @click="sendToGalaxy('scores')">Scores</Button>&nbsp;
+          <template v-if="countColumns.length != 0">
+            <Button class="p-button-outlined p-button-sm" @click="sendToGalaxy('counts')">Counts</Button>&nbsp;
+          </template>
+          <Button class="p-button-outlined p-button-sm" @click="sendToGalaxy('mappedVariants')">Mapped Variants</Button>&nbsp;
         </div>
         <div v-if="item.abstractText">
           <div class="mave-score-set-section-title">Abstract</div>
@@ -314,11 +313,10 @@ export default {
       type: String,
       required: true
     },
-    GALAXY_URL: String,
-    tool_id: String,
-    sendToGalaxy: String
+    galaxyUrl: String,
+    toolId: String,
+    requestFromGalaxy: String
   },
-  // mounted() { console.log('Props:', this.GALAXY_URL, this.tool_id, this.sendToGalaxy); },
   data: () => ({
     scores: null,
     scoresTable: [],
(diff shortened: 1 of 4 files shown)


src/components/screens/ScoreSetView.vue
```diff
@@ -303,22 +309,23 @@ export default {
...
-
   props: {
     itemId: {
       type: String,
       required: true
-    }
+    },
+    GALAXY_URL: String,
+    tool_id: String,
+    sendToGalaxy: String
```
ashsny, line 319: any reason for the mix of ALL_CAPS, snake_case, and camelCase here?

ashsny: let's avoid having sendToGalaxy as a prop and sendGalaxy as a function.

function should probably be called sendToGalaxy, name of prop is up to you

ashsny (commented):
Thanks for the contribution! Could you record a little demo/take screenshots of how this works? I use https://recordit.co/ on OS X.

PlushZ:
https://github.com/VariantEffect/mavedb-ui/assets/55543056/24c8aafa-9daa-4b88-8424-e38f41363a96

ashsny:
changes LGTM, I will ask @jstone-uw to look today and then we should be good to merge.

jstone-uw:
It looks like Galaxy will open the MaveDB home screen in the user's browser with some URL query parameters, which will get stored in Vuex state and later used by ScoreSetView. Is that right? And does this happen in an iframe or in a new browser tab? (Sorry, I couldn't see the screencast for some reason.)

**Comments about the use of Vuex state**
- I’d suggest we be more specific about the expected variables, instead of just caching everything that may be in route.query and trying to pass it all later as props to ScoreSetView. This isn’t essential, but it seems to me it would be better to name the parameters explicitly in the code. Judging from ScoreSetView's new props, we’re just looking for query parameters named `galaxyUrl`, `toolId`, and `requestFromGalaxy`.
- As long as we’re storing these in Vuex state, why not just have ScoreSetView retrieve them from the Vuex state, instead of having the router pass them in as props?
- I think we’ll migrate other state from Vuex to Pinia soon. It’ll be simple enough to migrate Galaxy-related state at that point, so no need to change this now.

**Vue-router integration and persistence across reloads**

MaveDB's vue-router is currently configured in hash mode, where the route path and query parameters appear inside what's actually the main URL's fragment (#) section. So, for instance, the URL for search results for the KCNQ4 gene is
https://mavedb.org/#/search?search=KCNQ4.

The alternative is HTML5 mode, where the Vue route and query parameters appear in the main URL: https://mavedb.org/search?search=KCNQ4

See https://router.vuejs.org/guide/essentials/history-mode for a discussion of these modes. We can support HTML5 mode on CloudFront if we want by setting up a custom response to 404 Not Found errors that treats index.html as a fallback page.

Unless we change the vue-router history mode, I think Galaxy will need to send the user to something like https://mavedb.org/#/?galaxyUrl=X&toolID=Y&requestFromGalaxy=Z.

_Options_
1. Keep using hash mode, and have Galaxy send a URL like the one above. If the user navigates to another page, the query params will disappear from the URL. If the user then refreshes, they'll also be lost from Vuex state.
2. Keep using hash mode, but have Galaxy pass the query params in the main URL. This would look like https://mavedb.org/?galaxyUrl=X&toolID=Y&requestFromGalaxy=Z#/. We would need to change the mechanism for extracting parameters from the URL, but the parameters would stay there throughout the user's session, and they would even be passed to new browser tabs/windows that the user might open.
3. Switch to HTML5 mode. The user'e experience would be the same as option 1.

For this use case, option 2 seems preferable. However, I think we might want to switch to HTML5 mode to make URLs prettier, and this would be a good time to make the decision.

PlushZ:
> It looks like Galaxy will open the MaveDB home screen in the user's browser with some URL query parameters, which will get stored in Vuex state and later used by ScoreSetView. Is that right?

Yes.

> And does this happen in an iframe or in a new browser tab?

The MaveDB content opens in the same tab as the user is currently viewing (neither iframe nor new tab)

> * I’d suggest we be more specific about the expected variables, instead of just caching everything that may be in route.query and trying to pass it all later as props to ScoreSetView. This isn’t essential, but it seems to me it would be better to name the parameters explicitly in the code. Judging from ScoreSetView's new props, we’re just looking for query parameters named `galaxyUrl`, `toolId`, and `requestFromGalaxy`.

makes sense

> * As long as we’re storing these in Vuex state, why not just have ScoreSetView retrieve them from the Vuex state, instead of having the router pass them in as props?

also makes sense

> Unless we change the vue-router history mode, I think Galaxy will need to send the user to something like https://mavedb.org/#/?galaxyUrl=X&toolID=Y&requestFromGalaxy=Z.

Now Galaxy does this

jstone-uw:
Thanks! The changes look good to me.

@ashsny and I are leaning toward switching to HTML5 mode in the Vue router. We can probably make this change in production within the next week. I thought I'd mention it since this will affect the URL that Galaxy uses for passing parameters. I don't think any changes will be needed to this PR.

ashsny:
Merged into release-2023.5.0.

PlushZ:
Thank you so much! Please ping me once you switch to HTML5 mode and I check if I need to change anything on the Galaxy side.

PR #89, opened 2023-12-14
Change author query to param translation to use a list.

### LINKED ISSUE #87: Author search causes 422s.
Steps to reproduce:
1. go to https://mavedb.org/#/search
2. click 'publication filters' then any author
3. refresh page
(produces 422 error and blank search results)

same result can be had from just linking to a search author page (i.e. https://mavedb.org/#/search?publication-authors=Starita,+Lea+M)

ashsny:
Fixes #87.


jstone-uw (approved):
Tested and confirmed that it resolves the issue.

PR #91, opened 2023-12-15
Add types for the API using openapi-typescript and instructions to re-generate them.

ashsny:
The local/development version of these instructions relies on https://github.com/VariantEffect/mavedb-api/pull/118. This workflow is obviously not ideal, but exists in its current complexity only until a merge of the `mavedb-api` and `mavedb-ui` repos is effected (this merge will be needed for sensible integration tests that are automatically run in CI).

Of course, generating from the live version of the API is currently possible without this complexity, so I have added instructions for that as well.


jstone-uw (approved):
Looks good to me. The operation names are pretty unwieldy, but they make sense, and most of the type names for requests and responses are clean. I wonder if we can do something on the API side to clean up Body_upload_score_set_variant_data_api_v1_score_sets__urn__variants_data_post, but maybe not since it's a multipart request.

PR #96, opened 2023-12-22
Refactor #1 in preparation for typescriptifying SearchView.vue.

ashsny:
This change eliminates a bunch of repeated code in parsing of scoresets to create filter options and extracts lodash in favor of native JS APIs.

### DIFF
--- a/src/components/screens/SearchView.vue
+++ b/src/components/screens/SearchView.vue
@@ -56,7 +56,9 @@ function countScoreSetMetadata(scoreSets, scoreSetMetadataFn) {
   if (!scoreSets.length) {
     return []
   }
-  const values = scoreSets.map(scoreSetMetadataFn).flat();
+
+  // Filter out empty string values.
+  const values = scoreSets.map(scoreSetMetadataFn).flat().filter((item) => !!item);
   const frequencies = values.reduce((counts, item) => {
     counts.set(item, (counts.get(item) || 0) + 1)
     return counts
@@ -72,7 +74,7 @@ function countPublicationMetadata(scoreSets, publicationMetadataFn) {
     const secondary = scoreSet.secondaryPublicationIdentifiers.map(publicationMetadataFn).flat()
 
     // Use a Set to eliminate duplicate values, then transform it back into an Array.
-    return [...new Set<string>(primary.concat(secondary))]
+    return [...new Set(primary.concat(secondary))]
   })
 }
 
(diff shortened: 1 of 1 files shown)


src/components/screens/SearchView.vue
```diff
@@ -80,99 +103,31 @@ export default {
...
     targetAccessionFilterOptions: function() {
-      if (this.publishedScoreSets.length > 0) {
-        const values = this.publishedScoreSets.map((s) => _.concat(s.targetGenes.map((t) => _.get(t, 'targetAccession.accession', 'Sequence Based Scoresets'))).flat()).flat()
-        const valueFrequencies = _.countBy(values)
-        return _.sortBy(_.keys(valueFrequencies)).map((value) => ({value, badge: valueFrequencies[value]}))
-      } else {
-        return null
-      }
+      return countTargetGeneMetadata(this.publishedScoreSets,
+        (targetGene) => targetGene.targetAccession?.accession || 'Sequence Based Scoresets')
```
bencap, line 114: This was a problem with the original logic that I am only noticing now, but creating this `Sequence Based Scoresets` (and similarly the `Accession Based Scoresets`) option will end up sending a non-sensical filter request to the API:
```
{"text":null,"targets":null,"targetOrganismNames":null,"targetAccessions":["Sequence Based Scoresets"],"targetTypes":null,"authors":null,"databases":null,"journals":null}
```

It might not make sense to handle in a refactor branch, especially because it may require API changes as well, but we should consider sending a different payload than the filter label for this sort of an option. Alternatively, maybe it makes more sense to not include a filter option for sequence based stuff in the accession filter menu at all.

bencap (commented):
Love how much code duplication this gets rid of. Left a couple comments, but the one that requires logic changes might make more sense to handle separately from this refactor.

src/components/screens/SearchView.vue
ashsny: For now, let's not display them in the list. We can potentially work out something different in a separate issue/PR (but I think I actually like them not being represented in the filter list here).

src/components/screens/SearchView.vue
```diff
@@ -53,6 +52,30 @@ import TabPanel from 'primevue/tabpanel'
...
+function countTargetGeneMetadata(scoreSets, geneMetadataFn) {
+  return countScoreSetMetadata(scoreSets, (scoreSet) => [...new Set(scoreSet.targetGenes.map(geneMetadataFn))])
+}
+function countPublicationMetadata(scoreSets, publicationMetadataFn) {
+  return countScoreSetMetadata(scoreSets, (scoreSet) => {
+    const primary = scoreSet.primaryPublicationIdentifiers.map(publicationMetadataFn).flat()
+    const secondary = scoreSet.secondaryPublicationIdentifiers.map(publicationMetadataFn).flat()
+
+    // Use a Set to eliminate duplicate values, then transform it back into an Array.
+    return [...new Set<string>(primary.concat(secondary))]
```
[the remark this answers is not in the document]
ashsny: Oops, not sure how that got through. Fixed.

bencap (approved):

PR #97, opened 2023-12-28
Change from comma-separated query param to repeated query params (refactor 2 in preparation for Typescript).

ashsny:
The previous state (comma-separated) was troublesome with authors, which always include commas, which led to a bunch of fragile special casing (see https://github.com/VariantEffect/mavedb-ui/issues/87) and the inability to handle multiple author filters. Alternatives considered were URL encoding everything and using a non-comma separator (using some unreserved char like _, -, or ~). Vue router documentation was not clear w.r.t. the behavior of URL encoding (lots of references to automatic encoding as in [here](https://router.vuejs.org/guide/essentials/named-routes.html#Named-Routes) but our commas weren't getting encoded) and this seemed preferable to non-comma separation. 

With this change, URLs for multiple authors look like `search?publication-author=Adamovich,+Aleksandra+I&publication-author=Shendure,+Jay` (note the change from plural to singular, which is more accurate now) which are handled quite nicely by the vue router as arrays. There's a little bit of careful code around parsing, but that's mostly to satisfy typescript.

### DIFF
--- a/src/components/screens/SearchView.vue
+++ b/src/components/screens/SearchView.vue
@@ -77,6 +77,15 @@ function countPublicationMetadata(scoreSets, publicationMetadataFn) {
     return [...new Set(primary.concat(secondary))]
   })
 }
+function extractQueryParam(content) {
+  // If there are multiple values, they will be stored as multiple identical query params (i.e. ?a=1&b=2&a=3) and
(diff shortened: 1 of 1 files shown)


bencap (approved):
Looks good!

PR #101, opened 2024-01-11
Use `modelValue` in place of `inputTextValue` for External Target Gene Identifiers

### LINKED ISSUE #99: External Gene Identifiers don't persist after the Autocomplete loses focus.
On version `release-2023.5.0`, it is difficult to add External Gene Identifiers because any changes made in the autocomplete field do not persist if the user hits enter or clicks out of the autocomplete.

To reproduce:
- Start the dev server on branch `release-2023.5.0`.
- Create a new score-set and try typing in one of the External Gene Identifier autocomplete fields

These fields will not hold your data and it will reset with each key stroke. Copy/pasting data in will also erase it rapidly. Based on my investigation, it looks like an issue with the `inputTextValue` property of Vue refs that the External Gene Identifiers rely on to persist data changes.

bencap:
Fixes #99

### DIFF
--- a/src/components/screens/ScoreSetEditor.vue
+++ b/src/components/screens/ScoreSetEditor.vue
@@ -1003,11 +1003,11 @@ export default {
 
     acceptNewDoiIdentifier: function () {
       const input = this.$refs.doiIdentifiersInput
-      const searchText = (input.inputTextValue || '').trim()
+      const searchText = (input.modelValue || '').trim()
       if (validateDoi(searchText)) {
         const doi = normalizeDoi(searchText)
         this.doiIdentifiers = _.uniqBy([...this.doiIdentifiers, { identifier: doi }])
-        input.inputTextValue = null
+        input.modelValue = null
 
         // Clear the text input.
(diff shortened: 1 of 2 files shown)


src/components/screens/ScoreSetEditor.vue
```diff
@@ -1098,7 +1098,6 @@ export default {
     searchTargetGeneIdentifiers: function (dbName, event) {
       const searchText = (event.query || '').trim()
       if (searchText.length > 0) {
-        this.targetGene.externalIdentifiers[dbName].identifier = null
```
ashsny, line 1101: just making sure this is intentional as it seems different from the other changes

bencap: Yeah it was. With that line in there any time a user types the text box gets cleared and the text will only reappear once the Autocomplete loses its focus. I deleted it so the text never gets hidden in the text box, but there isn't any change in functionality.

ashsny:
approved

PR #104, opened 2024-01-16
Add /signed-in.html and /signed-out.html that were in webpack version to vite/rollup.

ashsny:
Unfortunately, we cannot do these without extra files, as per the Vite documentation here: https://vitejs.dev/guide/build#multi-page-app.

Relevant excerpt: Note that for HTML files, Vite ignores the name given to the entry in the rollupOptions.input object and instead respects the resolved id of the file when generating the HTML asset in the dist folder. This ensures a consistent structure with the way the dev server works.

### DIFF
--- a/signed-in.html
+++ b/signed-in.html
@@ -1,0 +1,18 @@
+<!DOCTYPE html>
+<html lang="">
+  <head>
+    <meta charset="utf-8">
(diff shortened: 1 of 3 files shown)


ashsny:
Actually, this seems to break sign-in with `npm run dev` (although not with `npm run preview`). Given that sign-in and -out seem to work for both without this change, I think I'll abandon this.

Issue #73, opened 2023-11-24
Looking up Target Accession/Transcript Identifier is slow 


nickzoic:
I discuss this in https://github.com/VariantEffect/mavedb-api/issues/112 but it could be either a front-end or a back-end fix, or maybe both.

bencap:
Completed via #105.

PR #105, opened 2024-01-17
Based on Ashley's geneName codes, modify Assembly dropdown codes together

EstelleDa:
Try to solve https://github.com/VariantEffect/mavedb-api/issues/112 and https://github.com/VariantEffect/mavedb-ui/issues/73

Based on https://github.com/VariantEffect/mavedb-ui/tree/ashsny/target-accession-lookups branch, add some more modifications.

### DIFF
--- a/src/components/screens/ScoreSetEditor.vue
+++ b/src/components/screens/ScoreSetEditor.vue
@@ -1267,7 +1267,6 @@ export default {
         }
         this.referenceGenome = this.item.referenceGenome
         this.assembly = this.item.assembly
-        this.geneName = this.item.geneName
         this.targetGenes = this.item.targetGenes
         this.extraMetadata = this.item.extraMetadata
       } else {
--- a/src/components/screens/ScoreSetEditor.vue
+++ b/src/components/screens/ScoreSetEditor.vue
@@ -349,7 +349,7 @@
                       <div class="field field-columns">
                         <div class="field-column">
                           <span class="p-float-label">
-                            <InputText v-model="targetGene.name" :id="$scopedId('input-targetGeneName')" />
+                            <InputText v-model="targetGene.name" :id="$scopedId('input-targetGeneName')" style="width: 100%"/>
                             <label :for="$scopedId('input-targetGene')">Target gene name</label>
                           </span>
                         </div>
@@ -357,7 +357,7 @@
                           <span class="p-float-label">
                             <!-- Assembly is the reference genome property in coordinate cases -->
                             <Dropdown v-model="assembly" :id="$scopedId('input-targetGeneAssembly')" :options="assemblies"
-                              optionGroupLabel="type" optionGroupChildren="assemblies">
+                              optionGroupLabel="type" optionGroupChildren="assemblies" style="width: 100%">
                               <template #optiongroup="slotProps">
                                 <div class="flex align-items-center dropdown-option-group">
                                   <div>{{ slotProps.option.type }}</div>
@@ -372,7 +372,7 @@
                           <span class="p-float-label">
                             <Dropdown v-model="geneName" :id="$scopedId('input-targetGeneGeneNames')"
                               :options="geneNamesAsObject" optionLabel="name" filter
-                              :virtualScrollerOptions="{ itemSize: 50 }" @change="autofillGeneName" />
+                              :virtualScrollerOptions="{ itemSize: 50 }" @change="autofillGeneName" style="width: 100%"/>
                             <label :for="$scopedId('input-targetGeneAssembly')">HGNC Name</label>
                           </span>
                         </div>
@@ -706,7 +706,9 @@ export default {
     },
     referenceGenome: null,
     assembly: null,
+    assemblySuggestions: [],
     geneName: null,
+    geneNameAccessionSuggestions: [],
     targetOptions: ["Assembly", "HGNC"],
     targetAutocomplete: 'HGNC',
     extraMetadata: {},
@@ -781,7 +783,7 @@ export default {
       else {
         return this.geneNames.map(name => ({ name }))
       }
-    }
+    },
   },
 
   watch: {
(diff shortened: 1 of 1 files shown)


src/components/screens/ScoreSetEditor.vue
```diff
@@ -846,6 +848,22 @@ export default {
           this.licenseId = this.defaultLicenseId
         }
       }
+    },
+    geneName: {
+      handler: async function(newValue, oldValue) {
+        if (newValue == oldValue) {
+          return;
+        }
+        this.geneNameAccessionSuggestions = await this.fetchTargetAccessionsByGene(this.geneName.name.trim())
```
ashsny, line 857: this will error if this.geneName is null (which it may be if the dropdown has no value selected) why not the this.geneNameDropdownValue = this.geneName?.name.trim() and then use that here and below?

src/components/screens/ScoreSetEditor.vue
```diff
@@ -897,16 +915,14 @@ export default {
     },
 
     fetchTargetAccessions: async function (event) {
-      if (this.targetAutocomplete == 'Assembly') {
-        const assembly = this.assembly?.trim() || null
-        if (assembly) {
-          this.targetGeneAccessionSuggestions = await this.fetchTargetAccessionsByAssembly(assembly)
+      if (this.targetAutocomplete == 'Assembly' ) {
```
ashsny, line 918: I would probably write a watcher for this.targetautocomplete to do the switching here.

src/components/screens/ScoreSetEditor.vue
```diff
@@ -846,6 +848,22 @@ export default {
...
+        }
+        this.geneNameAccessionSuggestions = await this.fetchTargetAccessionsByGene(this.geneName.name.trim())
+      }
+    },
+    assembly: {
+      handler: async function(newValue, oldValue) {
+        if (newValue == oldValue) {
+          return;
+        }
+        this.assemblySuggestions = await this.fetchTargetAccessionsByAssembly(this.assembly.trim())
```
ashsny, line 865: why ditch the question mark here?

src/components/screens/ScoreSetEditor.vue
EstelleDa: Yes, you're correct. It will throw error if I click the Clear Target button.

src/components/screens/ScoreSetEditor.vue
```diff
@@ -846,7 +850,27 @@ export default {
...
       }
-    }
+    },
+    geneName: {
+      handler: async function(newValue, oldValue) {
+        if (newValue == oldValue) {
+          return;
+        }
+        this.geneNameDropdownValue = this.geneName?.name || null
+        this.geneNameAccessionSuggestions = await this.fetchTargetAccessionsByGene(this.geneNameDropdownValue)
```
ashsny, line 860: should we add a similar if clause here?

src/components/screens/ScoreSetEditor.vue
```diff
@@ -846,7 +850,27 @@ export default {
...
+      }
+    },
+    assembly: {
+      handler: async function(newValue, oldValue) {
+        if (newValue == oldValue) {
+          return;
+        }
+        this.assemblyDropdownValue = this.assembly?.trim() || null
+        if (this.assemblyDropdownValue){
```
ashsny, line 869: there should be a space between ) and {

src/components/screens/ScoreSetEditor.vue
ashsny, line 1: add new data fields to resetTarget

change reference in addTarget from `this.geneName?.name || null` to `this.geneNameDropdownValue`

src/components/screens/ScoreSetEditor.vue
```diff
@@ -1303,8 +1337,8 @@ export default {
         delete this.targetGene.targetAccession;
       }
       else if (this.assembly || this.geneName) {
-        this.targetGene.targetAccession.assembly = this.assembly || null
-        this.targetGene.targetAccession.gene = this.geneName?.name || null // Name property on string object array
+        this.targetGene.targetAccession.assembly = this.assemblyDropdownValue
```
ashsny, line 1340: This is not equivalent to the previous code (not sure if the previous code was correct) as assemblyDropdownValue has .trim() called on it. Are you sure the previous version was incorrect here?

src/components/screens/ScoreSetEditor.vue
ashsny: ok to skip

src/components/screens/ScoreSetEditor.vue
```diff
@@ -1243,6 +1267,7 @@ export default {
         }
         this.referenceGenome = this.item.referenceGenome
         this.assembly = this.item.assembly
+        this.geneName = this.item.geneName
```
ashsny, line 1270: don't believe there is a geneName on this.item. as the dropdown/autocomplete part we're using is only used during scoreset creation and this is only invoked when fetching a completed scoreset, probably best to skip this.

src/components/screens/ScoreSetEditor.vue
EstelleDa: These two lines of codes are in addTarget function and getting value of assembly or geneName, that's why I think we still need .trim() to remove leading and trailing whitespaces from the assembly and geneName. Based on that I think we can use this.assemblyDropdownValue and this.geneNameDropdownValue at here either.

src/components/screens/ScoreSetEditor.vue
EstelleDa: Yes, you're correct.

src/components/screens/ScoreSetEditor.vue
EstelleDa: I did some trials. I think Assembly/HGNC only accepts dropdown menu values. Do I had better change this part back to the previous code? I'm not sure whether in this case we need to care leading and trailing whitespaces or not.

ashsny: fine to leave it as this.assemblyDropdownValue, I don't think the lack of trim() has any advantage.

ashsny (approved):
looks good to go to me!

Issue #109, opened 2024-01-19
Clear button throws error Uncaught (in promise) TypeError: Cannot read properties of null (reading 'UniProt')


EstelleDa:
If only typing something in short description of creating a new score set and clicking Clear button, Clear button won't work and will throw an error Uncaught (in promise) TypeError: Cannot read properties of null (reading 'UniProt') from 
```
'targetGene.externalIdentifiers': {
      deep: true,
      handler: function (newValue) {
        // If an identifier has been set, set the offset to 0 by default.
        for (const dbName of externalGeneDatabases) {
          // Error is from this line
          if (newValue[dbName]?.identifier?.identifier != null && newValue[dbName]?.offset == null) {
            this.targetGene.externalIdentifiers[dbName].offset = 0
          }
        }
      }
    },
    
```
But clicking Clear button again, it'll work.

EstelleDa:
In Mavedb website, after clicking the Clear button in creating a new score set, Target Sequence and Genomic Coordinates totally disappear.

bencap:
Completed via #113.

PR #113, opened 2024-01-25
Solve #109 problem

EstelleDa:


src/components/screens/ScoreSetEditor.vue
```diff
@@ -1282,7 +1284,27 @@ export default {
         this.secondaryPublicationIdentifiers = []
         this.publicationIdentifiers = []
         this.dataUsagePolicy = null
-        this.targetGene = null
+        this.existingTargetGene = null
+        this.targetGene = {
```
ashsny, line 1288: this is redundant with l686 where target gene is created. I think you want to add a helper function emptyTargetGene outside the component to wrap this code and then find all the similar code and replace it with this.

src/components/screens/ScoreSetEditor.vue
```diff
@@ -790,13 +790,15 @@ export default {
     'targetGene.externalIdentifiers': {
       deep: true,
       handler: function (newValue) {
-        // If an identifier has been set, set the offset to 0 by default.
-        for (const dbName of externalGeneDatabases) {
-          if (newValue[dbName]?.identifier?.identifier != null && newValue[dbName]?.offset == null) {
-            this.targetGene.externalIdentifiers[dbName].offset = 0
+        if (newValue) {
```
ashsny, line 793: instead of wrapping the whole function in `if (newValue) {`
just add 

```
if (!newValue) {
  return
}
```

src/components/screens/ScoreSetEditor.vue
```diff
@@ -1291,7 +1313,6 @@ export default {
         this.geneName = null
         this.geneNameAccessionSuggestions = []
         this.geneNameDropdownValue =  null
-        this.targetGenes = []
```
ashsny, line 1294: why is this deleted? what is the effect of not clearing this on clear()?

src/components/screens/ScoreSetEditor.vue
EstelleDa: Good idea.

src/components/screens/ScoreSetEditor.vue
EstelleDa: It's repeated. Another `this.targetGenes = []` is in line 1308.

src/components/screens/ScoreSetEditor.vue
```diff
@@ -1282,22 +1285,21 @@ export default {
...
+      this.assembly = null
+      this.assemblySuggestions = []
+      this.assemblyDropdownValue = null
+      this.existingTargetGene = null
+      this.geneName = null
+      this.geneNameAccessionSuggestions = []
+      this.geneNameDropdownValue =  null
       this.fileCleared('targetGeneTargetSequenceSequenceFile')
+      this.referenceGenome = null
       this.targetGene = {
```
ashsny, line 1303: please also fix redundancy with l686 by adding helper function returning this target gene

ashsny: it looks like you're not yet using the helper function for the initialization at l686.

src/components/screens/ScoreSetEditor.vue
ashsny: this.targetGenes is no longer getting cleared in the 'New score set' case when clear is clicked due to this deletion. not sure what you mean by it's repeated here, I don't see it on l1308 before or after your changes.

ashsny:
I'll finish this up as I want it to go in with the frontend push today.

src/components/screens/ScoreSetEditor.vue
EstelleDa: Maybe the repeated codes were from my other branch.

src/components/screens/ScoreSetEditor.vue
EstelleDa: Thanks Ashley! Now I know what is the helper function.

PR #123, opened 2024-02-20
Home Page Highlights View for Score Set and Target Aggregate Statistics

bencap:
Adds a Highlight View component to display simple counting statistics of key Target, Score set, and/or Experiment model properties. Counting statistics are generally displayed as leaderboard style data, where property values that are associated with more Score sets, Experiments and/or Targets are displayed first. Where appropriate, data that is more readily displayed as proportional is displayed with charts.

For Score sets and Experiments, this may include Keywords, Publication Identifiers, DOI Identifiers, and/or Raw Read Identifiers:
Screenshot 2024-02-20 at 3 19 48 PM

For Targets, these fields include Accession and Gene names: 
Screenshot 2024-02-20 at 3 22 45 PM

and Ensembl, RefSeq, and UniProt identifiers:
Screenshot 2024-02-20 at 3 23 26 PM

Certain target properties, such as Assemblies, Organisms, References, and Gene Categories make more sense to display proportionally. However if desired we could also display them as a leaderboard, or just drop them from the homepage. Hovering over a segment will display its legend:
Screenshot 2024-02-20 at 3 34 13 PM

Chart segments and identifiers are all selectable and link to the internal MaveDB search page, where associated Score sets can be displayed and explored. Publication identifiers link to the internal publication identifier pages. Where appropriate, external URLs are linked to the appropriate data provider.

This component is re-used on the homepage to display both Score set and Target gene summary information, which may be useful to visitors wanting to engage with the site in a more exploratory fashion:
Screenshot 2024-02-20 at 3 28 44 PM

In the future, it would be useful to provide additional context around identifier based fields (such as DOI, RefSeq, UniProt, and Ensembl Identifiers) so that users can understand more about the field value than just the identifier name. It will also be useful to add properties to these routes as they are added to Score sets and Targets, such as the new Taxonomy information that will represent an upgrade over the currently used Keywords. It should be simple to add these properties as additional configuration entries in addition to adding fetcher functions to appropriately format statistic API responses.

These changes also include a config entry for the application URL. As we add more internal links, it will be useful to ensure links clicked within a development environment also open with the development URL.

### DIFF
--- a/.env.dev
+++ b/.env.dev
@@ -1,2 +1,3 @@
 VITE_API_URL=http://localhost:8002/api/v1
+VITE_APP_URL=https://localhost:8081
 VITE_SITE_TITLE=MaveDB (local API)
--- a/src/components/common/HighlightsView.vue
+++ b/src/components/common/HighlightsView.vue
@@ -357,9 +357,6 @@ export default defineComponent({
           window.open(`${config.appBaseUrl}/#/search?${model}=${Object.keys(data)[element[0].index]}`)
         },
         plugins: {
-          colorschemes: {
-            scheme: 'brewer.Paired12'
-          },
           legend: {
             display: false
           },
@@ -378,7 +375,7 @@ export default defineComponent({
         datasets: [
           {
             data: entries.map((e) => { return e[1] }),
-            // Color pallete for pie charts.
+            // Colors for pie charts; Colors palette from https://sashamaps.net/docs/resources/20-colors/.
             backgroundColor: ['#3f51b5', '#e6194b', '#3cb44b', '#ffe119', '#f032e6', '#bcf60c', '#fabebe', '#008080', '#e6beff', '#f58231', '#911eb4', '#4363d8', '#46f0f0', '#9a6324', '#fffac8', '#800000', '#aaffc3', '#808000', '#ffd8b1', '#808080', '#ffffff', '#000000']
           }
         ]
--- a/src/components/common/HighlightsView.vue
+++ b/src/components/common/HighlightsView.vue
@@ -30,20 +30,20 @@
                   </template>
                   <!-- Handle url columns separately, so that we can fill in any missing DB data -->
                   <template v-else-if="this.field == 'uniprot-identifier' && col.field == 'url'" #body="slotProps">
-                    <a
-                      :href="slotProps.data[col.field] ? `${slotProps.data[col.field]}` : `http://purl.uniprot.org/uniprot/${slotProps.data.identifier}`" target="_blank">{{
+                    <a :href="slotProps.data[col.field] ? `${slotProps.data[col.field]}` : `http://purl.uniprot.org/uniprot/${slotProps.data.identifier}`"
+                      target="_blank">{{
                       slotProps.data.url ? slotProps.data.url :
                       `http://purl.uniprot.org/uniprot/${slotProps.data.identifier}` }}</a>
                   </template>
                   <template v-else-if="this.field == 'refseq-identifier' && col.field == 'url'" #body="slotProps">
-                    <a
-                      :href="slotProps.data[col.field] ? `${slotProps.data[col.field]}` : `http://www.ncbi.nlm.nih.gov/entrez/viewer.fcgi?val=${slotProps.data.identifier}`" target="_blank">{{
+                    <a :href="slotProps.data[col.field] ? `${slotProps.data[col.field]}` : `http://www.ncbi.nlm.nih.gov/entrez/viewer.fcgi?val=${slotProps.data.identifier}`"
+                      target="_blank">{{
                       slotProps.data.url ? slotProps.data.url :
                       `http://www.ncbi.nlm.nih.gov/entrez/viewer.fcgi?val=${slotProps.data.identifier}` }}</a>
                   </template>
                   <template v-else-if="this.field == 'ensembl-identifier' && col.field == 'url'" #body="slotProps">
-                    <a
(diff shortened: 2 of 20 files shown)


.env.dev
```diff
@@ -1,2 +1,3 @@
 VITE_API_URL=http://localhost:8002/api/v1
-VITE_SITE_TITLE=MaveDB (local API)
```
ashsny, line 2: this is a little awkward, because currently this is used as a switch to switch "which API backend am I using while developing?" so when I'm developing only the frontend locally, I run `MODE=live npm run dev` and don't have to run a local backend.

I think we do want a mode where these links point to localhost but the prod api is used instead of a local api. I think the best solution is to have 3 modes:
1. local api and app (dev)
2. prod api, local app (prodapi or some better name)
3. prod api, prod app (live)

and update the readme accordingly.

sorry for the hassle!

src/components/common/HighlightsView.vue
```diff
@@ -0,0 +1,615 @@
...
+      targetGeneReferenceFieldCounts: targetGeneReferenceStatistic.item,
+      targetGeneOrganismFieldCounts: targetGeneOrganismStatistic.item,
+
+      field: field,
+      activeTabIndex: activeTabIndex,
+      loading: loading
+    }
+  },
+
+  computed: {
```
ashsny, line 293: necessary?

src/components/common/HighlightsView.vue
```diff
@@ -0,0 +1,615 @@
...
+    },
+
+    statisticsDictToChartData: function (stats) {
+      let entries = Object.entries(stats)
+      return {
+        labels: entries.map(e => { return e[0] }),
+        datasets: [
+          {
+            data: entries.map(e => { return e[1] }),
+            backgroundColor: ['#3f51b5', '#e6194b', '#3cb44b', '#ffe119', '#f032e6', '#bcf60c', '#fabebe', '#008080', '#e6beff', '#f58231', '#911eb4', '#4363d8', '#46f0f0', '#9a6324', '#fffac8', '#800000', '#aaffc3', '#808000', '#ffd8b1', '#808080', '#ffffff', '#000000']
```
ashsny, line 385: are these just the brewer colors? maybe add a commet to this effect

src/components/common/HighlightsView.vue
```diff
@@ -0,0 +1,615 @@
...
+            text: title
+          },
+        }
+      }
+    },
+
+    statisticsDictToChartData: function (stats) {
+      let entries = Object.entries(stats)
+      return {
+        labels: entries.map(e => { return e[0] }),
```
ashsny, line 381: add parens around e here to be consistent

src/components/common/HighlightsView.vue
```diff
@@ -0,0 +1,615 @@
...
+
+        return identifiers
+      }
+
+      // Target fields
+      if (this.field == 'accession' || this.field == 'gene') {
+        return this.countsToLeaderboard(this.dataForField)
+      }
+      if (this.field == 'uniprot-identifier') {
+        return (await loadUniprotIdentifiers()).sort((a, b) => { a.count - b.count })
```
ashsny, line 465: extract sorting fn into local var

src/lib/item-types.js
```diff
@@ -128,6 +128,28 @@ const itemTypes = {
       }
     }
   },
+  'target-gene-statistics': {
```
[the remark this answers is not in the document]
bencap: Yeah, I am still using the item types for the pie charts while using the API routes on leaderboard style data.

src/components/screens/HomeScreen.vue
```diff
@@ -1,7 +1,7 @@
 <template>
```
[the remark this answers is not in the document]
bencap: The only reason I left it was because I like the way it visually adds another column. Not sure what would go in its place, but agreed it's redundant given what information is on the stats view.

src/components/common/HighlightsView.vue
```diff
@@ -378,10 +374,11 @@ export default defineComponent({
     statisticsDictToChartData: function (stats) {
       let entries = Object.entries(stats)
       return {
-        labels: entries.map(e => { return e[0] }),
+        labels: entries.map((e) => { return e[0] }),
         datasets: [
           {
-            data: entries.map(e => { return e[1] }),
+            data: entries.map((e) => { return e[1] }),
+            // Color pallete for pie charts.
```
ashsny, line 381: s/pallete/palette. I guess I meant "what palette is this? where did you get it?" not "what are these colors for?"

src/components/common/HighlightsView.vue
```diff
@@ -503,29 +466,29 @@ export default defineComponent({
...
+          `${config.apiBaseUrl}/${field}/search`,
           { text: identifier },
           {
             headers: {
               accept: 'application/json'
             }
           }
         )
-        // TODO catch errors in response
+        // TODO mavedb-ui#130 catch errors in response
```
ashsny, line 488: does github like this formatting? for some reason I had it in my head that we should use TODO(#130) for issues in the same repo

bencap: Yeah that format is better, idk why I thought I needed the repo when obviously it would be in the UI repo lol. Fixed

src/components/common/HighlightsView.vue
bencap: I'll add a comment for it. I wanted to just use some brewer palette but they topped out at 12 colors and I think we already have more than 12 reference genomes (which I'm sort of worried about as more get added). I ended up just googling for differentiated colors and got these from here: https://sashamaps.net/docs/resources/20-colors/. There is probably a better way than just a list of colors though.

ashsny: list of colors is fine as long as you put where you got it

PR #124, opened 2024-02-26
Quick fix for UI issues with experiment/score set editor autocompletes.

ashsny:
This is only a temporary fix. A durable fix will unify DOI/publication ID searching so that there is only one, full-featured way to search and add publications (probably including search by author/title?). 

This drops autocomplete for DOIs and Raw Read IDs, which was at present only coming from values already in the database.  Primevue's changes mean we can no longer search the database and then use `@keyup.enter` to trigger adding values not in the database, so I chose to drop search instead to maintain the functionality to add new values. I think this is okay on a temporary basis because we don't expect to repeat those values often (therefore using values already in the database), and these are unique IDs, not mildly ambiguous categories/strings. A future change will have to come up with a solution for raw read ids, if we want to add back that functionality.

### DIFF
--- a/src/components/screens/ExperimentEditor.vue
+++ b/src/components/screens/ExperimentEditor.vue
@@ -62,18 +62,21 @@
               </div>
               <div class="field">
                 <span class="p-float-label">
-                  <AutoComplete
+                  <Chips
                       ref="doiIdentifiersInput"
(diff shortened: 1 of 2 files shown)


bencap (approved):
Code looks good to me. Was able to add some DOIs / Raw Reads locally with no issues.

PR #126, opened 2024-02-28
Fixed: Documentation link is broken in AWS deployment.

jstone-uw:
CloudFront doesn't allow us easily to specify a default document except at the website root. (This can be accomplished using a Lambda function.) Instead, we'll just include index.html in the link.


ashsny:
should we merge into 2023.5.1 instead? currently identical to main

jstone-uw:
Thanks, that's what I meant to do.

PR #134, opened 2024-03-02
Add a simple histogram of scores to ScoreSetView.

ashsny:
In a mind-bogglingly hot take, there are no axis labels currently. I'm not sure if they're useful given they're redundant with the title and tooltip, but they should probably be there anyway.

See screen recording: https://recordit.co/ACOKD9UYMu


jstone-uw:
The code looked good to me; my only comment is that I wonder whether it makes sense to show the distribution at the top of the page. If we haven't already, let's get input from investigators before merging.

ashsny:
Added axis labels: 
image

PR #144, opened 2024-03-22
Use ProcessingState to Display Variant Processing Information on Unpublished Score Sets

bencap:
Uses the `processingState` score set property and associated `processingErrors` property to display information about the state of the variant processing job for Score Sets. Errors in the job are now displayed in a more friendly table, replacing the toast notification that may have contained information about hundred of validation failures. These UI elements are only displayed on unpublished score sets.

When score set variant data has been submitted for processing
Screenshot 2024-03-22 at 3 55 30 PM

When score set variant data processing has failed
Screenshot 2024-03-22 at 3 56 22 PM

There is also a table with variant validation errors included in the variant section
Screenshot 2024-03-22 at 4 00 02 PM

For target accessions, errors often look slightly different
Screenshot 2024-03-22 at 4 00 58 PM

For successfully processed variants the existing table will display a selection of the first 10 variants and a success banner will display at the top of the page.
Screenshot 2024-03-22 at 4 03 21 PM

### DIFF
--- a/src/components/screens/ScoreSetView.vue
+++ b/src/components/screens/ScoreSetView.vue
@@ -2,6 +2,28 @@
   <DefaultLayout>
     <div v-if="item" class="mave-score-set">
       <div class="mave-1000px-col">
+        <div v-if="!item.publishedDate" class="variant-processing-status">
+          <div v-if="item.processingState == 'success'">
+            <Message severity="success">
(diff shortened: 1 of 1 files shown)


jstone-uw (approved):
I don't have any changes to suggest, and the error reporting will be a big improvement. I'm not sure yet whether we'll want to release this before or after 2024.0.0; let's discuss before merging.

Issue #146, opened 2024-03-25
Hide the heatmap if there is nothing to display.


jstone-uw:
Some score sets contain no single missense/nonsense variants, so that there is nothing to display on a heat map except, perhaps, wild-type alleles. In this case the heatmap should not be shown.

jstone-uw:
Released in version 2024.0.0 on 2024-04-15.

PR #150, opened 2024-04-02
Do not show heatmap unless there are some display-able variants.

ashsny:
Also, delete unused showSimpleVariants[Stacked]Heatmap.


ashsny:
Fixes #146

PR #160, opened 2024-04-12
Stop HighlightsView pie charts from re-rendering when tab is switched.

ashsny:
Make everything computed properties - chart.js options/data change handler was triggered on each tab change for extremely unclear reasons (presumably a perverse interaction with Vue reactivity). However, moving everything into a computed property (which is honestly probably the correct thing to do from Vue's perspective) no longer triggers this handler (presumably it would be called if the underlying data changed).

merging this one into newTaxonomy as opposed to release to avoid triggering a merge conflict. they will become one shortly in any case.

### DIFF
--- a/src/components/common/HighlightsView.vue
+++ b/src/components/common/HighlightsView.vue
@@ -57,22 +57,15 @@
             </div>
           </TabPanel>
         </TabView>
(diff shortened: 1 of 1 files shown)


bencap (commented):
Awesome 🙏 -- This bothered me for ages when I originally wrote this.

bencap (approved):

PR #170, opened 2024-04-15
Sort and create an 'others' category for charts. Re-order gene highlights to show accession-based stats last as these are mostly empty.

ashsny:


ashsny:
Screenshot 2024-04-15 at 2 24 43 PM

bencap (approved):
Looks good. I'm going to open an issue about the NULL organisms so we don't lose track of that.

Issue #135, opened 2024-03-05
Improved login mechanism


jstone-uw:
**Issues**
- Users report getting logged out and not being able to log in immediately.
- When using multiple tabs, users must log in separately in each tab.
- The login mechanism requires a popup window.

Some of these issues may inhere in the choice of ORCID for login, the use of OpenID Connect (OIDC) as a mechanism, or the particular OIDC library we're using now.

bencap:
Released in https://github.com/VariantEffect/mavedb-api/releases/tag/v2024.1.0

PR #176, opened 2024-04-16
Better ORCID authorization

ashsny:
Jeremy's original PR description follows:

Addresses #135.

The code for managing ORCID login sessions has been replaced in order to address three issues:

- Users report getting logged out and not being able to log in immediately.
- When using multiple tabs, users must log in separately in each tab.
- The login mechanism requires a popup window.

The rationale for this change is described in lib/orcid.ts:

> "Three-legged OAuth" is used here. We originally tried using third-party libaries -- primarily vuex-oidc, which in
> turn uses oidc-client or oidc-client-ts for the actual OAuth2/OIDC authentication process. We ran into issues with
> all OAuth2 and OIDC authentication flows:
> 
> - With OIDC's implicit flow, the client does not need to manage a client secret; this is good for single-page
> applications whose source code is visible. This flow is supported by vuex-oidc v3 via oidc-client, though it is no
> no longer supported by the supported branches of these projects. However, the access token has a limited lifespan,
> and ORCID does not seem to support token renewal. (It does not return a renewal token, and its renew endpoint has
> an X-Frame-Options header that prohibits use in an iframe, so in any case we cannot renew silently.) So the user is
> logged out after 10 minutes.
> - The preferred option would be OIDC's authorization code flow with PKCE, which is more secure than the implicit flow
> and still does not require a client secret. However, ORCID does not support PKCE. Plans to support it seem to have
> stalled. (See https://github.com/ORCID/ORCID-Source/issues/5977.)
> - So we have opted to use three-legged authentication, which requires participation of the MaveDB API. In this flow,
> the browser client first obtains a code when the user signs into ORCID and grants permissions. It sends this code
> to the MaveDB API, which makes a second request to ORCID, this time exchanging the code and MaveDB's client secret
> for an access token and, in addition, the ID token (JWT) that we will subsequently use for authorization. These
> have a very long lifespan (which ORCID's OAuth2 service returns in the expires_in parameter, measured in seconds).
> Three-legged authentication is not supported by oidc-client or oidc-client-ts, so the present module implements
> it.

This change replaces the vue-oidc-client and oidc-client-ts packages with custom authentication code. The new logic mainly resides in three files:
- lib/orcid.ts (which replaces those packages and much of lib/auth.js)
- lib/auth.ts (authorization utilities, replacing part of lib/auth.js)
- store/modules/auth.ts (authorization/role management, replacing store/modules/auth.js)

In addition, there are two new Vue components implementing callback screens: components/screens/OidcCallback.vue and components/screens/OidcCallbackError.vue.

**Release notes**
- Because of the switch to three-legged OAuth, this change must be released in conjunction with a new API endpoint.
- This change also switches the Vue router's history manager from hash mode to HTML5 mode. This will provide a better user experience, but it means that old bookmarks will send users to the homepage.
- Because of the router change, serving applications such as CloudFront must be configured to serve index.html in response to unknown page requests. This is already the case in MaveDB's production environment.


src/composition/auth.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * This composable uses two pieces of code:
+ * - The ORCID OAuth2 authentication module, lib/orcid.ts, which handles the OAuth2 authentication flow and stores
+ *   credentials in localStorage; and
+ * - The authorization store module, store/modules/auth.ts, which checks the user's roles after login.
+ *
+ * The user's authentication state, profile and roles are exposed here, as are functions for logging in and out, so Vue
+ * components should not need to interact with the 
+ */
+import _ from 'lodash'
```
ashsny, line 12: doesn't look like this file uses lodash

src/lib/orcid.ts
```diff
@@ -0,0 +1,231 @@
...
+  idToken.value = newIdToken
+}
+
+/**
+ * Begin the sign-in process by sending the user to ORCID's sign-in page.
+ *
+ * We send the following parameters to ORCID:
+ * - client_id: MaveDB's OAuth2/OIDC client ID for ORCID, which is not a secret
+ * - redirect_uri: The URI to which the user should be redirected after sign-in
+ * - scope=openid: A valie indicating that we want an OpenID ID token
```
ashsny, line 96: s/valie/value

src/lib/orcid.ts
```diff
@@ -0,0 +1,231 @@
...
+  id: string
+  jti: string
+  sub: string
+}
+
+// Notice the final '/'.
+const appUrl = `${window.location.origin}/`
+
+interface AuthState {
+  /** A JWT identifying the current user. User details can be obtained by decoding it. */
```
ashsny, line 50: why a multi-line comment here?

src/lib/orcid.ts
```diff
@@ -0,0 +1,231 @@
...
+ *
+ * ORCID does not support full sign-out, so the user will remain logged into ORCID for any other applications that
+ * may try to authenticate (or may already be authenticated). For MaveDB, we require ORCID to present the login
+ */
+export function signOut() {
+  clearIdToken()
+}
+
+/*
+// This is what authentication looks like using vuex-oidc v3 and the implicit flow.
```
ashsny, line 155: do we need to preserve this? is it possible this library will ever work for us?

src/lib/orcid.ts
```diff
@@ -0,0 +1,231 @@
...
+export interface OidcUserProfileBase {
+  auth_time: number
+  family_name: string
+  given_name: string
+  id: string
+  jti: string
+  sub: string
+}
+
+// Notice the final '/'.
```
ashsny, line 46: not sure I understand the comment here - is this used other than with oidc-callback appended on?

ashsny (commented):
importing my comments from other PR

jstone-uw:
@ashsny This change involves switching the Vue router mode, so that for instance instead of https://mavedb.org/#/score-sets/{urn}, the URL will be https://mavedb.org/score-sets/{urn}. I think we all prefer this style anyway, and the move to AWS facilitated it.

Do you think that at this stage in adoption it's worth addressing the problem of broken bookmarks? One possibility is to detect a URL fragment that's a valid path in the application, navigate accordingly, and display a pop-up suggesting that the user update their bookmarks.

src/components/screens/ExperimentView.vue
```diff
@@ -178,36 +178,35 @@
...
 import DefaultLayout from '@/components/layout/DefaultLayout'
 import PageLoading from '@/components/common/PageLoading'
 import ItemNotFound from '@/components/common/ItemNotFound'
 import ProgressSpinner from 'primevue/progressspinner'
+import useAuth from '@/composition/auth'
 import useItem from '@/composition/item'
 import useFormatters from '@/composition/formatters'
-import axios from 'axios'
+import config from '@/config'
 import { oidc } from '@/lib/auth'
```
bencap, line 196: This `import config from '@/config'` isn't needed here (duplicated from line 184), and I believe the `import { oidc } from '@/lib/auth'` references a method removed by these changes. Both of these throw compiler errors when running `npm run dev`.

jstone-uw: Thanks, not sure how that got into the checked-in version. I've removed it.

src/lib/orcid.ts
jstone-uw: I went ahead and removed this unused code.

PR #178, opened 2024-04-19
Improved Admin User Handling

bencap:
Adds UI support for role based API keys and a selectable active user.

*All API keys shown were local and have since been regenerated*

Non-admin users may still generate and delete user API keys as before:
Screenshot 2024-04-19 at 11 42 10 AM
Screenshot 2024-04-19 at 11 42 43 AM

Users with assigned roles now have access to the creation of role based API keys:
Screenshot 2024-04-19 at 11 43 34 AM
Screenshot 2024-04-19 at 11 43 56 AM

Users with assigned roles may now select an acting role. This role is stored within the `auth` store and sent to the API under the `X-Active-Role` header. When making authenticated API requests, the user will have access to only the permissions granted by the selected role. When a user has roles selected that are non-default, a warning banner will display above the page noting that they are using non-default permissions for API requests. The selected acting role has no influence over the screens shown to the user, and is only used for requests to the server. If a user has no acting role selected, they will receive the union of permissions from all assigned roles.
Screenshot 2024-04-19 at 11 48 00 AM
Screenshot 2024-04-19 at 11 48 15 AM
Screenshot 2024-04-19 at 11 48 37 AM


bencap:
UI support for https://github.com/VariantEffect/mavedb-api/pull/184

src/components/layout/Toolbar.vue
```diff
@@ -1,4 +1,10 @@
 <template>
+  <div v-if="(activeRole && activeRole != 'default') || (!activeRole && roles.length > 0)">
+    <Message severity="warn">
+      When making API requests, you are currently acting as a user with the [{{ activeRole || roles.join(", ") }}] role(s). You may have additional permissions than normal. To change your
```
ashsny, line 4: should be `You may have additional permissions` or `You may have more permissions than normal` but not both :P

src/components/screens/SettingsScreen.vue
```diff
@@ -75,35 +133,45 @@ export default {
     }
   },
 
-  data: function() {
+  data: function () {
```
ashsny, line 136: any idea why this is adding spaces between function names and parens? seems abnormal

bencap: Maybe something with my auto formatter for Vue files. I'll look into it.

ashsny:
the only thing that's a bit confusing to me here is the radio button role selection interface for users.

IIUC, it defaults to 'default' and no admin permissions. then we have two other options: 'admin selected': admin permissions, and 'unset': all roles selected, effectively admin permissions.

am I right? do we need the unset option for now given only admin and default are available?

is there maybe a better string we could show in this interface than 'default' to express that the user is acting with normal-ish perms?

bencap:
@ashsny That's all correct, with the caveat that it starts as unset when the user initially logs in and gives them all roles. 

And yes, if a user has admin permissions and selects them, then that is the same thing as having unset permissions. The implementation now though is flexible in that if new roles are added to the backend (are we going to want a curator role in the near future or is that not necessary?), they'd be available to us immediately in the client.

I thought about `user` rather than `default` for these permissions too, but wasn't really sure there was a great single word descriptor for them. Happy to change it if you had any ideas.

bencap:
Yeah that seems fine to me. Thinking about it some more it actually probably makes sense to keep the user box checked no matter what. The way the backend is designed right now you will always have user permissions so long as you are signed in. And I think I like 'ordinary user'.

And yeah, this will affect all actions. I suppose I made that distinction because selecting a role doesn't affect what a user will see in the UI (for instance, admin users have access to an additional `users` screen that still appears in the toolbar even if they are acting as a non-admin user).

I'll add that link in as well.

bencap:
A couple things to notice now:
- Checkboxes (the ordinary user box is always checked, since you will always have that role so long as you are authenticated)
- Settings screen link and some tweaks to the language in the warning banner
Screenshot 2024-04-24 at 10 09 12 AM

ashsny (approved):

PR #181, opened 2024-04-23
Request User Emails for Certain Site Features

bencap:
Includes three main changes to better support required emails for certain site features

Display validation errors for email updates within the settings screen:
Screenshot 2024-04-23 at 5 00 43 PM

Adds toast notifications for generic API failures to the ExperimentEditor screen so that trying to add an experiment without a valid user email won't fail silently:
Screenshot 2024-04-23 at 5 01 37 PM

Adds an importable dialog prompt component to capture user emails. This component can be added to any screen that will require an email to submit data. If a user does not have an email address set on their user profile, the component will appear as an overlay:
Screenshot 2024-04-23 at 5 04 11 PM


bencap:
UI Changes for https://github.com/VariantEffect/mavedb-api/pull/186

src/components/common/EmailPrompt.vue
```diff
@@ -0,0 +1,135 @@
+<template>
+    <Dialog v-model:visible="visible" :closable="false" :header=title :style="{ width: '25rem' }">
+        <span class="p-text-secondary block mb-5">{{ dialog }}</span>
+        <div class="flex align-items-center gap-3 mb-3">
+            <label for="email" class="font-semibold w-6rem">Email</label>
+            <InputText v-model:model-value="email" id="email" class="flex-auto" />
+        </div>
+        <div><span v-if="emailValidationError" class="mave-field-error">{{ emailValidationError }}</span></div>
+        <div class="flex justify-content-end gap-2">
+            <Button type="button" label="Ignore" severity="secondary" @click="ignoreEmail"></Button>
```
ashsny, line 10:
```suggestion
            <Button type="button" label="Not now" severity="secondary" @click="ignoreEmail"></Button>
```

ashsny: not super sure on the wording here. any brilliant ideas?

bencap: Not really, the only other thing I can think of is something like `Maybe later`.

src/components/common/EmailPrompt.vue
```diff
@@ -0,0 +1,135 @@
...
+                else {
+                    this.emailValidationError = null
+                    this.user.email = email
+                    this.$toast.add({ severity: 'success', summary: `Your email was succesfully updated.`, life: 3000 })
+                    this.visible = false
+                }
+            }
+        },
+
+        ignoreEmail: async function() {
```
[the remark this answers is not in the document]
bencap: Yeah, maybe this should get reworked a little tbh. It saves the user so that their `is_first_login` value in the backend gets set to `False`. That ensures that the component will only show up once for their first login. 

It's kinda relying on a side-effect of the update user endpoint though... (and we probably also would prefer to have the email field in the endpoint schema to not be optional). Maybe a dedicated endpoint would be better rather than this?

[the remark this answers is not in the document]
ashsny: yeah I would have an api side mechanism to set the new column to true directly

Issue #139, opened 2024-03-13
ScoreSetHeatmap: deletions visualized in Ala row


PlushZ:
In score set heatmap:

1. If there is only deletion - score visualization appears in A (Ala) row
![screenshot1](https://github.com/VariantEffect/mavedb-ui/assets/55543056/5bed12d3-c9a7-4996-a648-d0193a926209)

2. If there are both substitution by A and deletion - both scores appear in A row 
![screenshot3](https://github.com/VariantEffect/mavedb-ui/assets/55543056/a94ff350-e17a-4382-b555-c400d554a975)
![screenshot2](https://github.com/VariantEffect/mavedb-ui/assets/55543056/4b5ac8d8-7419-4045-8465-6420cfd1e841)

3. If there is only substitution by A - score appears correctly in A row

[csv with scores set for screenshots](https://github.com/VariantEffect/mavedb-ui/files/14589173/Scores.Dadonaite.BA.1.Delta.ab-escape.-.LyCoV-1404.1.csv)

PlushZ:
scoreset from screenshots is saved here https://www.mavedb.org/#/score-sets/tmp:540c9d47-9e0c-4c56-839c-3b1249a957df

PR #183, opened 2024-04-29
Fix heatmap bug in representing deletions.

jstone-uw:
This fixes issue #139, in which it was reported that deletions appear in the top row of the heatmap, normally reserved for substitutions by Ala.

In addition:
- Some refactoring, TypeScript migration, and JSDoc creation has also been done.
- The heatmap's y-axis tick mark labels have been made more readable.

@PlushZ provided a useful example in score set [tmp:540c9d47-9e0c-4c56-839c-3b1249a957df](https://www.mavedb.org/#/score-sets/tmp:540c9d47-9e0c-4c56-839c-3b1249a957df). To test the bug fix using this example, you may change the ownership of that score set in a development instance of MaveDB.

The heatmap for this score set should now look like this:
heatmap-tmp-540c9d47-9e0c-4c56-839c-3b1249a957df

The heatmap for score set urn:mavedb:00000001-a-1 is a good check against regressions, and it should look like this:
heatmap-urn-mavedb-00000001-a-1


src/components/ScoreSetHeatmap.vue
```diff
@@ -318,14 +319,24 @@ export default {
...
-              .style('font-size', 15)
+              .attr('class', 'mave-heatmap-y-axis-tick-labels')
               .call(
                 d3.axisLeft(yScale)
                     .tickSize(0)
-                    .tickFormat(n => AMINO_ACIDS_BY_HYDROPHILIA[20 - n])
-                    // .tickFormat(n => String.fromCharCode('A'.charCodeAt(0) + n - 1))
+                    .tickFormat((n) => {
+                      // Get the row's amino acid code or variation symbol.
+                      const label = HEATMAP_ROWS[HEATMAP_ROWS.length - 1 - n]
```
ashsny, line 328: repeating this logic to reverse the HEATMAP_ROWS gets very confusing to the reader.

I think there are a few options here. 

1) reverse heatmap_rows
2) use a d3 scalePoint as in this notebook to combine the flipping and translation of letter to axis position: https://observablehq.com/@mbostock/phases-of-the-moon
3) to eliminate the backwards iteration below, use a data structure for each heatmap_rows entry with style and label attributes that have defaults, so we have entries like {'symbol': '-', 'style': 'mave-heatmap-y-axis-tick-label-lg'} and {'symbol': '*', 'label': '\uff0a'}

src/lib/heatmap.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Given a MaveHGVS-pro amino acid code or code representing deletion, synonmyous variation, or stop codon, return the
+ * corresponding single-character code (which is the code used in our heatmap's y-axis).
+ *
+ * @param aaCodeOrChange A one- or three-character code representing an amino acid or the result of a variation at a
+ *   single locus in a protein sequence. If not an amino acid code, it should be a code representing synonymous
+ *   variation (=), stop codon (*), or deletion (- or del).
+ * @return The one-character code representing the same amino acid or change, or null if the input was not a supported
+ *   amino acid or change.
+ */
+export function singleLetterAminoAcidOrHgvsCode(aaCodeOrChange: string) {
```
ashsny, line 23: let's do explicit return types for ts in this project

PR #184, opened 2024-05-02
Variant Search

bencap:
Adds UI elements to the score set page to allow users to search for specific variants.

- Adds an auto complete search bar above the histogram and heatmap. Users may enter text into the auto complete and any matching HGVS strings will display below it for the user to select.
Screenshot 2024-05-03 at 9 33 48 AM

- The selected variant is now passed as a prop to the histogram and heatmap. When this prop is null, the histogram and heatmap behave as they do now. When a user selects a variant in the auto complete, the variant is passed to the visualizations as the external selection.
- The histogram will highlight the bin containing the external selection and display a tooltip with information about the bin.
- The heatmap will highlight both the variant square and stacked variant bar in addition to showing a tooltip with information about the variant.
Screenshot 2024-05-03 at 9 34 25 AM

- When you mouse over the histogram while a variant is selected, the tooltip will disappear and the tooltip related to your mouse over will display. When your mouse leaves the histogram focus, the selected variant tooltip will show up again.
- When you mouse over the heatmap while a variant is selected, an additional tooltip will display. If you mouse over the selected variant, it will disappear. When your mouse leaves the focus area for the heatmap, the additional tooltip will dissapear.
Screenshot 2024-05-03 at 9 35 14 AM

- Selecting a variant in the auto complete will scroll to the variant's position in the heatmap, if it is not already on the screen.
- Removed some unused code for clicks in the heatmap definition.
- Added an `emits` event for clicking a square in the heatmap. If you click a variant, the variant you click will be emitted to the parent component. This event is used by the score set view to change the variant selection.


src/components/ScoreSetHistogram.vue
```diff
@@ -57,6 +62,16 @@ export default {
       const bins = this.bins
       return bins
     },
+    selectedBin: function() {
+      const externalSelection = this.externalSelection
```
ashsny, line 66: I would do `self = this` instead of these two lines, then use self below

src/components/screens/ScoreSetView.vue
```diff
@@ -42,11 +42,39 @@
...
+                  <div v-if="variantNotNullOrNA(slotProps.option.hgvs_pro)">Protein variant: {{ slotProps.option.hgvs_pro }};&nbsp;</div>
+                  </div>
+              </template>
+            </AutoComplete>
+            <label :for="$scopedId('variant-search')">Search for a variant defined by this score set</label>
+            <Button icon="pi pi-times" severity="danger" aria-label="Clear" @click="selectedVariant = null" />
+          </span>
+        </div>
+        <div class="mave-score-set-histogram-pane">
+          <ScoreSetHistogram :scoreSet="item" :scores="scores" :externalSelection="typeof selectedVariant === 'object' ? selectedVariant : null" />
```
ashsny, line 73: having these `typeof`s here is a bit confusing. could this be a computed property?

src/lib/mave-hgvs.ts
```diff
@@ -36,3 +36,14 @@ export function parseSimpleProVariant(variant: string): SimpleProteinVariation |
     substitution: match[3]
   }
 }
+
+
+  /**
```
ashsny, line 41:
```suggestion
/**
```

src/components/screens/ScoreSetView.vue
```diff
@@ -42,11 +42,39 @@
...
+            >
+              <template #option="slotProps">
+                <div class="flex align-options-center">
+                  <div v-if="variantNotNullOrNA(slotProps.option.hgvs_nt)">Nucleotide variant: {{ slotProps.option.hgvs_nt }};&nbsp;</div>
+                  <div v-if="variantNotNullOrNA(slotProps.option.hgvs_splice)">Splice variant: {{ slotProps.option.hgvs_splice }};&nbsp;</div>
+                  <div v-if="variantNotNullOrNA(slotProps.option.hgvs_pro)">Protein variant: {{ slotProps.option.hgvs_pro }};&nbsp;</div>
+                  </div>
+              </template>
+            </AutoComplete>
+            <label :for="$scopedId('variant-search')">Search for a variant defined by this score set</label>
```
ashsny, line 68:
```suggestion
            <label :for="$scopedId('variant-search')">Search for a variant in this score set</label>
```

bencap:
Screenshot 2024-05-03 at 10 32 27 AM

ashsny:
thanks for adding the variant/score. consider this one approved

PR #191, opened 2024-05-15
Estelle/debug supersed score set

ashsny:
Frontend code for showing current version for superseded score sets by @EstelleDa .


EstelleDa:
release-2024.1.0 doesn't have correct deprecated score set codes. Current version of a score set of that branch doesn't work correctly.

PR #193, opened 2024-05-15
Allow Ter in the heatmap.

jstone-uw:
To date, the heatmap has only accepted the single-character code "*" for stop codons. This usage was present in some old score sets, but it is not valid MAVE-HGVS notation.

We will continue to support single-character notation for both amino acids and stop codons in the heatmap, but here we allow Ter as well. On the server side, single-character notation is not accepted, and the old score sets with invalid strings will be updated.

### DIFF
--- a/src/lib/heatmap.ts
+++ b/src/lib/heatmap.ts
@@ -3,7 +3,7 @@ import {AMINO_ACIDS, AMINO_ACIDS_BY_HYDROPHILIA} from './amino-acids.js'
 /** Codes used in the right part of a MaveHGVS-pro string representing a single variation in a protein sequence. */
 const MAVE_HGVS_PRO_CHANGE_CODES = [
   {codes: {single: '='}}, // Synonymous AA variant
-  {codes: {single: '*'}}, // Stop codon
+  {codes: {single: '*', triple: 'TER'}}, // Stop codon
   {codes: {single: '-', triple: 'DEL'}} // Deletion
 ]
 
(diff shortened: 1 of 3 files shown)


src/lib/heatmap.ts
```diff
@@ -3,7 +3,7 @@ import {AMINO_ACIDS, AMINO_ACIDS_BY_HYDROPHILIA} from './amino-acids.js'
 /** Codes used in the right part of a MaveHGVS-pro string representing a single variation in a protein sequence. */
 const MAVE_HGVS_PRO_CHANGE_CODES = [
   {codes: {single: '='}}, // Synonymous AA variant
-  {codes: {single: '*'}}, // Stop codon
+  {codes: {single: '*', triple: 'TER'}}, // Stop codon
```
bencap, line 6: It seems to me that technically, only the `=` matches the RegEx that validates simple protein variants. That seems fine though, since it seems like the mave-hgvs package wants to enforce the `TER` and `DEL` strings rather than the shorthand notation.

See: https://github.com/VariantEffect/mavedb-ui/blob/c1475cae81e2c7add1c049a10b0d8c26b7726f58/src/lib/mave-hgvs.ts#L19

bencap (approved):

src/lib/heatmap.ts
bencap: If we wanted to allow those single codes through, it would be pretty trivial to update the RegEx to something like `/^p\.([A-Za-z]{3})([0-9]+)([A-Za-z]{3}|=|\*|-)$/ `, but it doesn't seem like we necessarily want to support these shorthand codes.

jstone-uw: Thanks -- I updated this with a comment about allowing noncompliant MaveHGVS strings.

PR #194, opened 2024-05-15
Populate Target Properties on Autocomplete Selection

### LINKED ISSUE #192: Organism is not copied when duplicating an existing target.

bencap:
Sets additional helper properties on the score set on autocomplete selection of targets to allow the target to be saved directly. This will likely be a short lived change given planned updates to the score set editor.

Closes #192

### DIFF
--- a/src/components/screens/ScoreSetEditor.vue
+++ b/src/components/screens/ScoreSetEditor.vue
@@ -830,10 +830,8 @@ export default {
     existingTargetGene: {
       immediate: true,
       handler: function () {
-        console.log(this.existingTargetGene)
         if (_.isObject(this.existingTargetGene)) {
           // _.cloneDeep is needed because the target gene has been frozen.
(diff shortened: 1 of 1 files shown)


jstone-uw:
The taxonomy change looks good to me. I agree this part of the code is messy, but that's not new, and it will likely get refactored soon.

I wondered why we need to set other properties like `sequenceType`, `sequence`, `label`, and `accession` on the whole component. ScoreSetEditor doesn't seem to have these reactive properties defined (whereas it does have `taxonomy` and `assembly`), so I don't think this has an effect. But I might be missing something.

bencap:
Yeah, you're right. It's a little silly that the reactiveness for target sequences and target accessions works differently; I suppose I just assumed they worked the same. I'll push something up.

bencap:
Rolled into #195.

PR #195, opened 2024-05-16
Wizard-ify new score set screen.

ashsny:
For our initial base here, we'll use primevue's stepper/stepperpanel components. These only came out in Feb '24, so we have to do a long-delayed primevue version upgrade. This was delayed as it required fixing at least 3 resulting bugs, which I've done: header buttons not working as the MenuItem `to` property had been deprecated, badge numbers were not centered, and the main search page search field icon was not displaying correctly. I've fixed those bugs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "mavedb-ui",
-  "version": "2023.5.1",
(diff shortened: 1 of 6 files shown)


ashsny:
superseded by #197

PR #240, opened 2024-07-29
Debug creating a new score set fail if users choose superseded score set

EstelleDa:


bencap:
I made a mistake with the release branch, so merged the main branch into this one and force pushed so that we were up to date.

bencap (approved):
Thanks for taking care of this @EstelleDa, looks good. I'm going to make a change for this on the backend as well (unless you would like to) so we can make sure API users don't also encounter this.

bencap:
Gotcha thanks for looking, the validation error is probably sufficient. I'll deploy this tomorrow AM so that the user who is blocked by it is able to continue.

EstelleDa:
Thanks Ben! I'll let him know.

PR #243, opened 2024-08-02
Estelle/controlled keywords

EstelleDa:
Finished one.


src/components/screens/ExperimentEditor.vue
```diff
@@ -541,7 +752,7 @@ export default {
       }
 
       const editedItem = _.merge({}, this.item || {}, editedFields)
-
+      console.log(editedItem)
```
bencap, line 755: Can remove

src/components/screens/ExperimentView.vue
```diff
@@ -84,11 +84,32 @@
...
-            <a v-for="(keyword, i) of item.keywords" :key="i" :href="`${config.appBaseUrl}/search?search=${keyword}`"><Chip :label="keyword"/></a>
+            <li v-for="(keyword, index) in item.keywords" :key="index">
+              {{ keyword.keyword.key }}
+              <!--Present local database keyword description-->
+              <i class="pi pi-info-circle" style="color: green; cursor: pointer;" @click="showDialog(index)"/>
+              <Dialog v-model:visible="dialogVisible[index]" modal :header="keyword.keyword.key" :style="{ width: '50vw' }" :breakpoints="{ '1199px': '75vw', '575px': '90vw' }">
+                <p class="m-0">
+                  {{ keyword.keyword.description }}
+                </p>
+              </Dialog> : <a :href="`https://www.mavedb.org/search/?keywords=${keyword.keyword.value}`">{{ keyword.keyword.value }}</a>
```
bencap, line 98: It doesn't seem like there is a mechanism in the score set search to search by keywords on associated experiments. We either want to add that or remove this link, since right now it will just link to a generic search when clicked.

bencap (approved):
Looks really nice, left a couple minor comments but if you want to address them with TODOs and backlog items I don't think they are blocking.

bencap:
Resolves VariantEffect/mavedb-api#40

src/components/screens/ExperimentEditor.vue
EstelleDa: Agree. I think I added it when I debug something and forgot it.

src/components/screens/ExperimentView.vue
EstelleDa: In search_score_set function, I added this part. My understanding for it is the associated score sets will be returned if users type keywords. However, it's online one so I can't test it and I thought just put it first and see how it's going. 
image

bencap: I think that will allow us to load the score sets matching our filter logic, but we would still need to add the filter for keywords and add the . 

So in `src/mavedb/view_models/search.py` we would need to add a keywords list to the search model:
```
class ScoreSetsSearch(BaseModel):
    published: Optional[bool]
    targets: Optional[list[str]]
    target_organism_names: Optional[list[str]]
    target_types: Optional[list[str]]
    target_accessions: Optional[list[str]]
    authors: Optional[list[str]]
    databases: Optional[list[str]]
    journals: Optional[list[str]]
    publication_identifiers: Optional[list[str]]
    keywords: Optional[list[str]]
    text: Optional[str]
```

And then in `src/mavedb/lib/score_sets.py` we'd have to add a block that queries the experiment keywords:
```
if search.keywords:
    query = query.filter(
        Experiment.keyword_objs.any(
            ControlledKeyword.value.in_(search.keywords)
        )
    )
```

I haven't tested that, but would be happy to look together if you wanted to add it in.

PR #245, opened 2024-08-12
Data set contributors

jstone-dev:
Creators of experiments and score sets may now assign them lists of contributors. Contributors must be looked up by ORCID ID, and the API returns the ORCID user's name. If a contributor has a MaveDB user account, that user is granted the same permissions on the experiment or score set as its creator.

- User interfaces for assigning contributors to an experiment or score set
- Experiment and score set views now display lists of contributors.
- Pinia and rest-client-vue are used here to manage ORCID ID lookups. We'll likely want to replace other uses of Vuex and useItems with these libraries.

This PR depends on the corresponding contributors PR on mavedb-api (https://github.com/VariantEffect/mavedb-api/pull/259).

### DIFF
--- a/src/components/screens/ExperimentEditor.vue
+++ b/src/components/screens/ExperimentEditor.vue
@@ -334,6 +334,7 @@ import useItem from '@/composition/item'
 import useItems from '@/composition/items'
 import config from '@/config'
 import {normalizeDoi, normalizePubmedId, normalizeRawRead, validateDoi, validatePubmedId, validateRawRead} from '@/lib/identifiers'
+import {ORCID_ID_REGEX} from '@/lib/orcid'
 import useFormatters from '@/composition/formatters'
 
 const KEYWORDS = [
@@ -509,7 +510,6 @@ export default {
 
   computed: {
     contributorSuggestions: function() {
-      console.log(this.contributorLookupResult)
       return this.suggestionsForAutocomplete(this.contributorLookupResult ? [this.contributorLookupResult] : [])
     },
     publicationIdentifierSuggestionsList: function() {
@@ -599,8 +599,7 @@ export default {
 
     lookupContributor: function(event) {
       const searchText = (event.query || '').trim()
-      console.log(searchText)
-      if (searchText.length > 0) {
+      if (searchText.length > 0 && ORCID_ID_REGEX.test(searchText)) {
         this.setContributorLookupId(searchText)
         this.setContributorLookupEnabled(true)
       } else {
--- a/src/components/screens/ExperimentEditor.vue
+++ b/src/components/screens/ExperimentEditor.vue
@@ -204,17 +204,15 @@
               </div>
               <div class="field">
                 <span class="p-float-label">
-                  <AutoComplete
-                      ref="contributorsInput"
-                      v-model="contributors"
-                      :id="$scopedId('input-contributors')"
-                      :multiple="true"
-                      placeholder="Type an ORCID ID"
-                      :suggestions="contributorSuggestions"
-                      @complete="lookupContributor"
-                      @keyup.escape="clearContributorSearch"
-                      @item-select="acceptNewContributor"
-                      option-label="orcidId"
+                  <Chips
+                    ref="contributorsInput"
+                    v-model="contributors"
+                    :id="$scopedId('input-contributors')"
+                    :addOnBlur="true"
+                    :allowDuplicate="false"
+                    placeholder="Type or paste ORCID IDs here."
+                    @add="newContributorsAdded"
+                    @keyup.escape="clearContributorSearch"
                   >
                     <template #chip="slotProps">
                       <div>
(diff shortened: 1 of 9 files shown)


src/components/screens/ScoreSetCreator.vue
```diff
@@ -326,6 +326,47 @@
                       </span>
                     </div>
                   </div>
+                  <div class="mavedb-wizard-row">
+                    <div class="mavedb-wizard-help">
+                      <label>
+                        An optional list of keywords that describe the experimental design of the score set. Score sets can be searched by keyword.
+                      </label>
+                      <div class="mavedb-help-small">
+                        Examples: SGE, barcode sequencing, MPRA, complementation
```
bencap, line 335: Contributors help text

src/components/screens/ScoreSetCreator.vue
```diff
@@ -326,6 +326,47 @@
...
+                      </div>
+                    </div>
+                    <div class="mavedb-wizard-content field">
+                      <span class="p-float-label">
+                        <AutoComplete
+                            ref="contributorsInput"
+                            v-model="contributors"
```
bencap, line 342: We should consider adding the contributors to the list of things auto-populated by experiment. Alan noted that most of the time are the same between the two items so it'd be nice for that to be the default. 

We'd just need to add something like
`this.contributors = event.value.contributors` in `populateExperimentMetadata`.

src/components/screens/ScoreSetEditor.vue
```diff
@@ -764,10 +813,14 @@
       metaAnalyzesScoreSetSuggestions: [],
       supersededScoreSetSuggestions: [],
       targetGeneAccessionSuggestions: [],
-      validationErrors: {},
+      validationErrors: {}
     }),
 
     computed: {
+      contributorSuggestions: function() {
+        console.log(this.contributorLookupResult)
```
bencap, line 821:
```suggestion
```

src/components/screens/ScoreSetCreator.vue
```diff
@@ -1002,6 +1061,10 @@ export default {
   }),
 
   computed: {
+    contributorSuggestions: function() {
+        console.log(this.contributorLookupResult)
```
bencap, line 1065:
```suggestion
```

src/components/screens/ScoreSetCreator.vue
```diff
@@ -1162,6 +1225,58 @@ export default {
...
+        let description = `${this.contributors[newIdx].firstName} ${this.contributors[newIdx].lastName}`
+        if (description.length == 1) {
+          description = this.contributors[newIdx].orcidId
+        }
+        this.$toast.add({severity:'warning', summary: `The ORCID user "${description}" is already associated with this experiment`, life: 3000})
+      }
+    },
+
+    clearContributorSearch: function() {
+      console.log('clear')
```
bencap, line 1255:
```suggestion
```

bencap (commented):
Thanks Jeremy, looks good overall. Left some minor comments-- I can approve once we take this off of draft, you may still have some local changes.

src/components/screens/ExperimentEditor.vue
```diff
@@ -215,6 +215,37 @@
...
                 <span v-if="validationErrors.methodText" class="mave-field-error">{{validationErrors.methodText}}</span>
               </div>
+              <div class="field">
+                <span class="p-float-label">
+                  <AutoComplete
+                      ref="contributorsInput"
+                      v-model="contributors"
+                      :id="$scopedId('input-contributors')"
+                      :multiple="true"
+                      placeholder="Type an ORCID ID"
```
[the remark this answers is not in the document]
jstone-dev: Same here. It doesn't seem to have to do with the floating label, which was my first guess, or with the use of slots. I don't mind the behavior in this case, especially with the help text to the left; but I'd like to know what's causing it.

src/components/screens/ExperimentEditor.vue
```diff
@@ -308,6 +356,10 @@ export default {
   }),
 
   computed: {
+    contributorSuggestions: function() {
+      console.log(this.contributorLookupResult)
```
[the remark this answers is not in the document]
jstone-dev: Done

src/components/screens/ExperimentEditor.vue
```diff
@@ -336,6 +388,57 @@ export default {
...
+    clearContributorSearch: function() {
+      // This could change with a new PrimeVue version.
+      const input = this.$refs.contributorsInput
+      input.$refs.focusInput.value = ''
+    },
+
+    lookupContributor: function(event) {
+      const searchText = (event.query || '').trim()
+      console.log(searchText)
+      if (searchText.length > 0) {
```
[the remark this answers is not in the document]
jstone-dev: Yep, thanks.

src/components/screens/ScoreSetCreator.vue
jstone-dev: It's populated now from either the experiment or the superseded score set.

src/lib/orcid.ts
```diff
@@ -43,6 +43,8 @@ export interface OidcUserProfileBase {
   sub: string
 }
 
+export const ORCID_ID_REGEX = /^\d{4}-\d{4}-\d{4}-\d{4}$/
```
bencap, line 46: ORCID IDs may also end with an `X`.

```suggestion
export const ORCID_ID_REGEX = /^\d{4}-\d{4}-\d{4}-(\d{4}|\d{3}X)$/
```

bencap (commented):
Looks nice, thanks for the Chips change. Once the ORCID Regex is changed we can merge this in.

In addition to the ORCID thing: The warning keyword for toast notifications is now `warn` not `warning` (unrelated but it seems like a lot of libraries are deprecating `warning`-- the python default logger is doing so as well). We lose message styling when we pass `warning` rather than `warn`.

src/lib/orcid.ts
jstone-dev: Fixed -- thanks for pointing this out.

bencap (approved):

PR #252, opened 2024-08-22
Add force selection to Taxonomy

EstelleDa:
If no force selection, invalid taxonomy can be added in target gene and raise an error in saving a new score set. 
https://github.com/VariantEffect/mavedb-api/issues/266


bencap (approved):
Let's merge this into 3.0 instead of 2.3, I don't think we will do another patch release before the 3.0 release we are planning next week.

EstelleDa:
> Let's merge this into 3.0 instead of 2.3, I don't think we will do another patch release before the 3.0 release we are planning next week.

Sure!

PR #251, opened 2024-08-19
Fixed: New Value Assigned to Variant Instances Constant

bencap:
Fixes #250

PR #255, opened 2024-08-26
SVG and PNG Downloads for Heatmap

bencap:
#216, plus rolls in changes from #251


src/components/ScoreSetHeatmap.vue
```diff
@@ -100,8 +101,10 @@ export default {
     wtVariants: function() {
       return this.wtAminoAcids ? this.prepareWtVariants(this.wtAminoAcids) : []
     },
-    showHeatmap: function() {
-      return this.simpleVariants && this.simpleVariants.length
+    heatmapVisible: function() {
+      const heatmapVisible = this.simpleVariants && this.simpleVariants.length
+      this.$emit('heatmapVisible', heatmapVisible)
```
jstone-dev, line 106: It's probably wise avoid emitting an event from the function that recalculates a computed property. It's a little more code, but I would suggest adding a watcher to emit the event (and maybe the watcher should check whether the value has really changed first).

Aside from the general notion that computed property functions shouldn't have side-effects, it's possible for this function to be called repeatedly with no change in the computed value.

src/components/ScoreSetHistogram.vue
```diff
@@ -1,4 +1,5 @@
 <template>
+  <!-- This ref name is coupled to parent elements; do not change it. -->
```
jstone-dev, line 2: Might be nicer to expose it as an explicit property. Some developers even discourage fetching subcomponent properties in Vue, but I think it's a reasonable thing to do. Digging into a subcomponent's $refs seems more fragile.

If I remember correctly, one way to do this is to store the ref in a property in mounted(). There may be more elegant ways; and anyway you may decide to leave it as-is.

jstone-dev:
The limitation of the export feature from the gastruloid site carries over here: In some browsers, clicking one of the download buttons just saves an SVG file instead of prompting the user for a location and image format (SVG or PNG). I think this is fine, though at some point we might want to look into it. It may be an effect of browser settings that we can't work around.

PR #263, opened 2024-09-05
Wizardification of creating experiment workflow

EstelleDa:
Add wizardification of creating experiment workflow. 
Modify editing experiment page.

### DIFF
--- a/src/components/screens/ExperimentCreator.vue
+++ b/src/components/screens/ExperimentCreator.vue
@@ -1,0 +1,1348 @@
+<template>
+  <EmailPrompt
+      dialog="You must add an email address to your account to create or edit an experiment. You can do so below, or on the 'Settings' page."
+      :isFirstLoginPrompt="false"
+  />
+  {{ experimentSetUrn }}
+  <DefaultLayout>
+    <div class="mave-experiment-editor">
+      <div v-if="itemStatus != 'NotLoaded'" class="mave-screen-title-bar">
+        <div class="mave-screen-title">Edit experiment {{ this.item.urn }}</div>
+        <div v-if="item" class="mave-screen-title-controls">
+          <Button @click="saveEditContent">Save changes</Button>
+          <Button @click="resetForm" class="p-button-help">Clear</Button>
+          <Button @click="viewItem" class="p-button-warning">Cancel</Button>
+        </div>
+      </div>
+      <div v-else class="mave-screen-title-bar">
+        <div class="mave-screen-title">Create a new experiment</div>
+        <div class="mave-screen-title-controls">
+          <Button @click="validateAndSave">Save</Button>
+          <Button @click="resetForm" class="p-button-help">Clear</Button>
+          <Button @click="backDashboard" class="p-button-warning">Cancel</Button>
+        </div>
+      </div>
+      <div class="mavedb-wizard">
+        <Stepper v-model:activeStep="activeWizardStep">
+          <StepperPanel>
+            <template #header="{index, clickCallback}">
+              <button class="p-stepper-action" :disabled="maxWizardStepEntered < index || maxWizardStepValidated < index - 1" role="tab" @click="clickCallback">
+                <span class="p-stepper-number">{{ index + 1 }}</span>
+                <span class="p-stepper-title">Experiment information</span>
+              </button>
+            </template>
+            <template #content="{nextCallback: showNextWizardStep}">
+              <div class="mavedb-wizard-form">
+                <div class="mavedb-wizard-form-content-background"></div>
+                <div class="mavedb-wizard-row">
+                  <div class="mavedb-wizard-content-pane">
+                    <Message severity="info">
+                      To add an experiment to a different set, please navigate to the experiment set first and click "Add experiment."
+                    </Message>
+                  </div>
+                </div>
+                <div class="mavedb-wizard-row">
+                  <div class="mavedb-wizard-help">
+                    <label>A short title for the experiment, to be displayed at the top of the experiment's own page.</label>
+                    <div class="mavedb-help-small">
+                      Examples: UBE2I yeast complementation, BRCA1 Y2H, PTEN VAMP-seq,
+                      Massively parallel functional dissection of ECR11 enhancer
+                    </div>
+                  </div>
+                  <div class="mavedb-wizard-content field">
+                    <span class="p-float-label">
+                      <InputText v-model="title" :id="$scopedId('input-title')"/>
+                      <label :for="$scopedId('input-title')">Title</label>
+                    </span>
+                    <span v-if="validationErrors.title" class="mave-field-error">{{ validationErrors.title }}</span>
+                  </div>
+                </div>
+                <div class="mavedb-wizard-row">
+                  <div class="mavedb-wizard-help">
+                    <label>
+                      A high-level description of the experiment in one or two sentences, to be displayed in search results.
+                    </label>
+                    <div class="mavedb-help-small">
+                      Example: A machine-learning imputed and refined Deep Mutational Scan of the human SUMO1 using functional complementation in yeast..
+                    </div>
+                  </div>
+                  <div class="mavedb-wizard-content field">
+                    <span class="p-float-label">
+                      <Textarea v-model="shortDescription" :id="$scopedId('input-shortDescription')" rows="4"/>
+                      <label :for="$scopedId('input-shortDescription')">Short description</label>
+                    </span>
+                    <span v-if="validationErrors.shortDescription" class="mave-field-error">{{
+                      validationErrors.shortDescription }}</span>
+                  </div>
+                </div>
+                <div class="mavedb-wizard-row">
+                  <div class="mavedb-wizard-help">
+                    <label>
+                      The motivation for and approach of the experiment.
+                    </label>
+                    <div class="mavedb-help-small">
+                      May be formatted using <a target="_blank" href="https://daringfireball.net/projects/markdown/syntax">Markdown</a>.
+                      The focus should be on the MAVE data, rather than the full research contribution, so use your judgement
+                      when deciding what details are relevant. It is common that experiments and score sets share the same abstract text if they are from the same study.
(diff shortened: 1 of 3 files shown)


src/components/screens/ExperimentCreator.vue
```diff
@@ -0,0 +1,1348 @@
...
+                <span class="p-stepper-title">Experiment information</span>
+              </button>
+            </template>
+            <template #content="{nextCallback: showNextWizardStep}">
+              <div class="mavedb-wizard-form">
+                <div class="mavedb-wizard-form-content-background"></div>
+                <div class="mavedb-wizard-row">
+                  <div class="mavedb-wizard-content-pane">
+                    <Message severity="info">
+                      To add an experiment to a different set, please navigate to the experiment set first and click "Add experiment."
```
bencap, line 40:
```suggestion
                      You are currently adding an experiment to a new experiment set. To add an experiment to an existing experiment set, navigate to the existing experiment set and click the "Add experiment" button.
```

src/components/screens/ExperimentCreator.vue
```diff
@@ -0,0 +1,1348 @@
...
+      <div v-else class="mave-screen-title-bar">
+        <div class="mave-screen-title">Create a new experiment</div>
+        <div class="mave-screen-title-controls">
+          <Button @click="validateAndSave">Save</Button>
+          <Button @click="resetForm" class="p-button-help">Clear</Button>
+          <Button @click="backDashboard" class="p-button-warning">Cancel</Button>
+        </div>
+      </div>
+      <div class="mavedb-wizard">
+        <Stepper v-model:activeStep="activeWizardStep">
```
bencap, line 26: I think it would be nice to have another step before the information page where the user can select the experiment set they will add their experiment to, sort of like what we have where users can select which experiment to add their score set to.

src/components/screens/ExperimentCreator.vue
```diff
@@ -0,0 +1,1348 @@
...
+                    <span v-if="validationErrors.methodText" class="mave-field-error">{{ validationErrors.methodText }}</span>
+                  </div>
+                </div>
+                <div class="mavedb-wizard-row">
+                  <div class="mavedb-wizard-help">
+                    <label>
+                      Digital Object Identifier
+                    </label>
+                    <div class="mavedb-help-small">
+                      Preprints or publications that are not indexed by PubMed can be included via the DOI field.
```
bencap, line 143: Let's make the help text for DOIs the same on this page as already exists on the score set page.

src/components/screens/ExperimentCreator.vue
```diff
@@ -0,0 +1,1348 @@
...
+                      <label :for="$scopedId('input-doiIdentifiers')">DOIs</label>
+                    </span>
+                    <span v-if="validationErrors.doiIdentifiers" class="mave-field-error">{{validationErrors.doiIdentifiers}}</span>
+                  </div>
+                </div>
+                <div v-if="itemStatus == 'NotLoaded' || this.item.private == true">
+                  <div class="mavedb-wizard-row">
+                    <div class="mavedb-wizard-help">
+                      <label>
+                        Contributors who may edit this experiment. Enter each contributor's
```
bencap, line 172: Let's make the help text for contributors the same on this page as already exists on the score set page.

src/components/screens/ExperimentCreator.vue
```diff
@@ -0,0 +1,1348 @@
...
+                        </Chips>
+                        <label :for="$scopedId('input-contributors')">Contributors</label>
+                      </span>
+                      <span v-if="validationErrors.contributors" class="mave-field-error">{{validationErrors.contributors}}</span>
+                    </div>
+                  </div>
+                  <div class="mavedb-wizard-row">
+                    <div class="mavedb-wizard-help">
+                      <label>
+                        Any publications associated with the experiment.
```
bencap, line 206: Let's make the help text for publication identifiers the same on this page as already exists on the score set page.

src/components/screens/ExperimentCreator.vue
```diff
@@ -0,0 +1,1348 @@
...
+            <template #header="{index, clickCallback}">
+              <button class="p-stepper-action" :disabled="maxWizardStepEntered < index || maxWizardStepValidated < index - 1" role="tab" @click="clickCallback">
+                <span class="p-stepper-number">{{ index + 1 }}</span>
+                <span class="p-stepper-title">Keywords</span>
+              </button>
+            </template>
+            <template #content="{prevCallback: showPreviousWizardStep}">
+              <div class="mavedb-wizard-row">
+                <div class="mavedb-wizard-help">
+                  <label>Choose related keywords(Optional).</label>
```
bencap, line 362:
```suggestion
                  <label>Choose related keywords (Optional).</label>
```

src/components/screens/ExperimentCreator.vue
```diff
@@ -0,0 +1,1348 @@
...
+                  <div class="mavedb-help-small">
+                    Experiments can be tagged with optional keywords.
+                    In a future release, the keyword vocabulary will become restricted and keyword selection will be mandatory. 
+                  </div>
+                </div>
+                <div class="mavedb-wizard-content field">
+                  <div class="keyword-editor">
+                    <div v-for="keyword in keywordData" :key="keyword.key">
```
bencap, line 370: I know you spent a lot of time on the keyword UI (and I think it looks good), but might it make sense to have a row for each of the keyword categories with help text on the left now that we are putting this in the wizard? Space was more limited in the previous upload flow than it is now. 

We could even have input switches like in the score set wizard that would help to indicate that these fields are optional. Users would select that yes they want to add a keyword for a specific vocabulary. I'm not wedded to that though, since it does add significantly more clicks.

src/components/screens/ExperimentCreator.vue
EstelleDa: That's what I considered before too. But experiment set only has an urn and I think it's a bad dropdown menu cause they're all too similar. In order to distinguish them, users still need to find the experiment set first. And not every new experiment belongs to an existing experiment set. Hence I think it's fine if they want to add it to an existing experiment set, then find the experiment set first. Otherwise, just creating a new experiment set for this experiment directly. That's why I didn't add this part.

src/components/screens/ExperimentCreator.vue
EstelleDa: I don't see the DOI in `ScoreSetCreator` page, but it's in `ScoreSetEditor` page. Do we miss it in `ScoreSetCreator` page?  
@afrubin Does creating a new score set have DOI? I remember we had it before.

src/components/screens/ExperimentCreator.vue
EstelleDa: Both of them are nearly the same, except score set is "Contributors who may edit this score set" and experiment is "Contributors who may edit this experiment". I think edit this experiment is more suitable except they cannot edit this experiment and only can edit related score set.

src/components/screens/ExperimentCreator.vue
EstelleDa: Both of them are nearly the same, but this one is for experiment so I changed the "score set" to "experiment" in the help text. I think it's better to say "Any publications associated with the experiment" rather than "Any publications associated with the score set" in creating new experiment page. The additional text is "Publications included in an experiment will also be displayed on their associated score set pages". It's from our document https://mavedb.org/docs/mavedb/upload_guide.html. I think it's helpful for users to understand what they can put in experiment so I added.

src/components/screens/ExperimentCreator.vue
bencap: Yeah, that makes sense. Thanks for the explanation

src/components/screens/ExperimentCreator.vue
bencap: Yeah it got dropped out somewhere, not sure why. I added it back in the target gene wizardification change with this help text:

```
                    <div class="mavedb-wizard-help">
                      <label :id="$scopedId('input-doiIdentifiers')">
                        The DOIs of any digital resources associated with the score set.
                      </label>
                      <div class="mavedb-help-small">
                        Please note: If you would like to associate publications with this score set via their DOI, please do not do so here.
                        Instead, use the publication identifiers field below.
                      </div>
                    </div>
```

src/components/screens/ExperimentCreator.vue
EstelleDa: No worries! This part was originally used in viewing experiment page. Ashley suggested me to add them in creating new experiment too so they're in here. I'll move them to the left help text. After confirming the final keywords list, all of them will be compulsory so I think we can simplify the optional effect first.

EstelleDa: Sorry, I think keep these description in small information is better because all of these keywords are in a for loop. If users choose either "endogenous locus library method" or "in vitro construct library method", new keyword options will pop out. The layout is very ugly and they don't match as a row orderly. We can adjust this part when we get the final list cause at that time the keywords' relationships may change.

bencap (approved):
Thanks Estelle, looks good! Just had one question about a duplicated line for you to look at.

src/components/screens/ExperimentEditor.vue
```diff
@@ -806,6 +786,24 @@ export default {
...
+          this.keywordKeys[key] = keywordObj ? keywordObj.keyword.value : null
+          this.keywordDescriptions[key] = keywordObj ? keywordObj.description : null
+        }
+        for (const k of KEYWORDS) {
+          setKeyword(k.key)
+        }
+      } else {
+        this.keywords = []
+        this.keywordKeys = _.fromPairs(KEYWORDS.map((keyword) => [keyword.key, null]))
+        this.keywordKeys = _.fromPairs(KEYWORDS.map((keyword) => [keyword.key, null]))
```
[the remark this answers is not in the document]
EstelleDa: I think so. I can't remember what happened to this line before. Maybe mistouch the keyboard and made it duplicate. It should be the keyword description.

PR #276, opened 2024-09-18
Score Range UI

bencap:
Accept score ranges during score set creation, editing, and viewing.

Default of no score ranges
Screenshot 2024-09-20 at 11 00 28 AM

Adding a single score range (not possible technically, since one abnormal and one normal is needed)
Screenshot 2024-09-20 at 11 01 03 AM

Users may add an arbitrary number of score ranges
Screenshot 2024-09-20 at 11 01 44 AM

### DIFF
--- a/src/components/screens/ScoreSetCreator.vue
+++ b/src/components/screens/ScoreSetCreator.vue
@@ -919,7 +919,7 @@
                   </div>
                   <div class="mavedb-wizard-content">
                     <span class="p-float-label">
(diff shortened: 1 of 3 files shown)


jstone-dev (commented):
I noticed a few things during manual testing:

- The WT score input doesn't currently accept decimal numbers, just integers.
- The "Remove" button placement seems odd, though I don't have a better suggestion. Maybe giving each score range an even stronger visual grouping would help; but I'm not very concerned about this, and I think we can revisit it if we like.
- I haven't been able to advance beyond the score range step, because validateScoreRanges never seems to return true.

Otherwise the code all looks good to me.

jstone-dev (approved):

PR #278, opened 2024-09-27
Heatmap Color Bar

bencap:
Fixes a bug where heatmap x-axis labels overlapped for n > 100

Adds a vertical color legend
Screenshot 2024-09-30 at 12 18 12 PM

Adds a distribution color bar to the bottom of the heatmap. NOTE: This element is not rendered in this change while we evaluate its utility.
Screenshot 2024-09-27 at 3 43 09 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,6 +56,7 @@
         "yup": "^1.3.3"
       },
       "devDependencies": {
+        "@types/d3": "^7.4.3",
         "@types/dom-to-image": "^2.6.7",
         "@types/lodash": "^4.17.0",
         "@types/papaparse": "^5.3.14",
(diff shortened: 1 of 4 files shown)


jstone-dev (commented):
The code looks good to me.

For some reason I think it might look better if the bar were shorter; maybe it's because the bottom rows of the heatmap are often empty. I'm not sure if the user needs quite so much detail, or if it would be enough to show a shorter bar since individual score values can be read from the heatmap itself. I'd defer to your judgment in any case.

I agree with holding back the horizontal bar, but I like the idea, and if users agree that it has some utility, then I think we just need to figure out what to call it / how to make its use clear.

bencap:
Screenshot 2024-10-10 at 3 02 42 PM
How this looks after the changes

PR #291, opened 2024-10-02
Solve any users can click edit or add button in experiment, score set and experiment set pages.

EstelleDa:


bencap (commented):
What if we added this as a function within the item store? Like a check permission type of deal so we only had to have it in one place? Note that I'm writing this snippet with the assumption we can pass in an action.

in `src/store/modules/item.js` within `actions`
```
    actionPermittedOnItem ({}, action) {
      try {
        // this response should be true to get authorization
        let response = await axios.get(`${config.apiBaseUrl}/experiment-sets/check-authorizations/${this.itemId}/action`)
        let authorized = response.data
      } catch (err) {
        console.log(`Error to get authorization:`, err)
        let authorized = False
      }
      return authorized
    }
```

Then we can just say `this.actionPermittedOnItem("add_experiment")` (or whatever action we are checking) and we can check the action is permitted on any item with the appropriately defined endpoint. 

@jstone-dev might know better than me if this is a reasonable use case of the item module. I'm a little hesitant to cache fetched permissions in the itemStore. If we are just using the permissions to decide what buttons to show to a user that seems fine, and then if permissions changed during the store lifetime they just will get an error when they try to invoke the action in question. I'm hesitant since we'd want to be careful though about doing client side filtering based on cached permission values, and I wonder if having these in the permission store makes that somewhat tempting when we shouldn't even have that possibility.

I think that if we don't want to use the itemStore for this sort of thing then your implementation looks good, we'd just need to update it for the suggested changes over in the API.

bencap:
Jeremy and I talked about this and think it's probably overcomplicating it, so I think this method looks good. Once we update the backend changes we can update these and then merge them together. I think the only thing might be to rename the functions to something more descriptive about the auth, like `checkUserAuthorization` or something.

bencap (approved):
Consider this approved, but let's wait to merge until we have the names of the routes finalized within https://github.com/VariantEffect/mavedb-api/pull/325. We might need to make a couple small changes here too.

PR #306, opened 2024-10-17
Include the URN in the Experiment dropdown when creating score sets

EstelleDa:
Experiments are ordered in title so the private and public are mixed together. We can change it in backend.

### DIFF
--- a/src/components/screens/ScoreSetCreator.vue
+++ b/src/components/screens/ScoreSetCreator.vue
@@ -95,25 +95,25 @@
                       </div>
                       <div v-else style="position: relative;">
                         <span class="p-float-label">
-                          <AutoComplete
+                          <Dropdown
                             ref="experimentInput"
                             v-model="experiment"
-                            dropdown
                             :id="$scopedId('input-experiment')"
-                            :suggestions="experimentSuggestionsList"
-                            field="title"
-                            forceSelection
-                            :multiple="false"
+                            filter optionLabel="title"
                             :options="editableExperiments"
+                            :virtualScrollerOptions="{ itemSize: 50 }"
                             @option-select="populateExperimentMetadata"
-                            @complete="searchExperiments"
-                            @keyup.escape="clearExperimentSearch"
                             style="width: 100%;"
                           >
-                            <template #item="slotProps">
-                              {{slotProps.item.urn}}: {{slotProps.item.title}}
+                            <template #option="slotProps">
+                              {{slotProps.option.urn}}: {{slotProps.option.title}}
                             </template>
-                          </AutoComplete>
+                            <template #empty>
+                              <div style="padding: 10px; text-align:center;">
+                                No experiments found.
+                              </div>
+                            </template>
+                          </Dropdown>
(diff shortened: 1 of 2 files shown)


src/components/screens/ScoreSetCreator.vue
```diff
@@ -95,15 +95,25 @@
                       </div>
                       <div v-else style="position: relative;">
                         <span class="p-float-label">
-                          <Dropdown
+                          <AutoComplete
+                            ref="experimentInput"
                             v-model="experiment"
+                            dropdown
                             :id="$scopedId('input-experiment')"
+                            :suggestions="experimentSuggestionsList"
```
bencap, line 103: I would suggest using `editableExperiments` here directly. The suggestions list is more for the autocomplete Vue component, and the dropdown is much better at handling empty lists. This will be useful for another comment below.

src/components/screens/ScoreSetCreator.vue
```diff
@@ -95,15 +95,25 @@
...
-                            @change="populateExperimentMetadata"
-                          />
+                            @option-select="populateExperimentMetadata"
+                            @complete="searchExperiments"
+                            @keyup.escape="clearExperimentSearch"
+                            style="width: 100%;"
+                          >
+                            <template #item="slotProps">
+                              {{slotProps.item.urn}}: {{slotProps.item.title}}
+                            </template>
```
bencap, line 115: If we use the `editableExperiments` property directly, it will let us use the `#empty` template to display a nice message when there are no matching experiments instead of just showing a colon. We can add this in addition to the `#item` template.

```
                            <template #empty>
                              <div style="padding: 10px; text-align:center;">
                                No experiments found.
                              </div>
                            </template>
```

src/components/screens/ScoreSetCreator.vue
```diff
@@ -95,15 +95,25 @@
...
+                          <AutoComplete
+                            ref="experimentInput"
                             v-model="experiment"
+                            dropdown
                             :id="$scopedId('input-experiment')"
+                            :suggestions="experimentSuggestionsList"
+                            field="title"
+                            forceSelection
+                            :multiple="false"
                             :options="editableExperiments"
```
bencap, line 107: I think another nice addition here would be to use the virtual scroller:
```
:virtualScrollerOptions="{ itemSize: 50 }"
```

Although it might not be super useful for me and you on staging, it will be nice for the users who have a huge amount of experiments since it will make them load much quicker.

bencap (commented):
Thanks Estelle, left a couple small comments about the Dropdown component but everything else looked good.

src/components/screens/ScoreSetCreator.vue
EstelleDa: I changed it to dropdown with filtering function. It looks better. Autocomplete has a small bug with the dropdown button.

bencap (approved):
Thanks for the changes Estelle, this looks really nice!

EstelleDa:
Thanks for giving me the feedback!

PR #308, opened 2024-10-22
Only Allow Users to Select from Active Licenses

bencap:
During score set creation, we now only display active licenses to the user for selection. Because some score sets may have licenses selected which are no longer active, these licenses may be retained if they exist on a score set, but are un-selectable otherwise. Once a user updates a score set to get rid of such licenses, they will no longer be selectable. In addition, a new warning is displayed while editing such score sets so users are notified they are using a deprecated license.

Coupled to https://github.com/VariantEffect/mavedb-api/pull/344

### DIFF
--- a/src/components/screens/ScoreSetCreator.vue
+++ b/src/components/screens/ScoreSetCreator.vue
@@ -322,11 +322,20 @@
                       </Message>
                     </div>
                   </div>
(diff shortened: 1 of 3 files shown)


jstone-dev (approved):
See my comment on https://github.com/VariantEffect/mavedb-api/pull/344. If you decide it's necessary to allow deprecated licenses on superseding score sets, then a UI change would be needed too. Otherwise, all good.

PR #310, opened 2024-10-24
Not showing meta-analysis experiments in creating score set dropdown menu.

EstelleDa:


bencap (commented):
My sense with this particular issue is that we should just filter the list of experiments returned to us by the API such that they exclude meta-analysis experiments. That way we don't need to make any API changes. The filter options here are a little wonky, since they don't actually serve any purpose at the moment other than denoting whether to show all experiments or just a user's experiments.

EstelleDa:
> My sense with this particular issue is that we should just filter the list of experiments returned to us by the API such that they exclude meta-analysis experiments. That way we don't need to make any API changes. The filter options here are a little wonky, since they don't actually serve any purpose at the moment other than denoting whether to show all experiments or just a user's experiments.

I changed the back-end codes too so that front end excludes the meta-analysis experiment directly. The main filter function is in https://github.com/VariantEffect/mavedb-api/pull/347

bencap:
> I changed the back-end codes too so that front end excludes the meta-analysis experiment directly. The main filter function is in [VariantEffect/mavedb-api#347](https://github.com/VariantEffect/mavedb-api/pull/347).

Yeah I guess that what I'm saying is that either we should make this query field hew closely to the score set structure (passing `meta_analyzes_score_set: false` will include only those score sets that are not meta-analysis score sets and to get only the users' current score sets we pass `user_id: <current_user_id>` instead of the weird current status quo where passing `some: value` is what does that), or we should just leave the query JSON as something we fix later on and do the filtration on the front-end.

src/lib/item-types.js
```diff
@@ -130,11 +130,6 @@ const itemTypes = {
       }
     }
   },
-  'experiment': {
-    name: 'experiment', // TODO Redundant, change this structure
-    restCollectionName: 'experiments',
-    primaryKey: 'urn'
-  },
```
bencap, line 137: We will still need this definition, since it is used for the experiment creator, editor, and viewer.

bencap (commented):
Thanks for this Estelle, the logic here looks good! We will just need to add the same logic to the ScoreSetEditor file, since that also needs a view of the users' editable experiments. Once we add that it'll be all set. Sorry for the delay in my review.

src/lib/item-types.js
EstelleDa: Yes, you're correct.

PR #313, opened 2024-10-28
Fixed: Edit Button not Appearing After Experiment Publication

bencap:
Now that we handle this button state by asking the backend whether a user has permission for a certain action, it doesn't make sense to base the div component's visibility off the publication date of the experiment. Make the div visible always, and control the button visibility conditionally based on backend responses.


EstelleDa (approved):
It looks great and works on my side.

PR #328, opened 2024-12-17
Contributor Supplied Score Thresholds

bencap:
UI features for https://github.com/VariantEffect/mavedb-api/pull/361.

- Adds a CalibrationTable injectable element
- Makes histogram shaded regions more generic
- Adds first-class logic for calibration and range formatting
- Adds CalibrationTable to Histogram display
- Eliminates score range hardcoding

### DIFF
--- a/src/components/CalibrationTable.vue
+++ b/src/components/CalibrationTable.vue
@@ -1,0 +1,77 @@
+<template>
+    <table v-if="evidenceStrengthsForDisplay" class="mave-calibrations-table">
+        <tr>
+            <th>Evidence Strength</th>
+            <th v-for="evidenceStrength of evidenceStrengthsForDisplay" :key="evidenceStrength"><b>{{ evidenceStrength }}</b></th>
+        </tr>
+        <tr>
+            <td>Likelihood Ratio+</td>
+            <td v-for="likelihoodRatio of positiveLikelihoodRatiosForDisplay" :key="likelihoodRatio">{{ likelihoodRatio }}</td>
+        </tr>
+        <tr>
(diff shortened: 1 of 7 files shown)


jstone-dev (commented):
I loaded Dan's sample calibrations for urn:mavedb:00000097-0-1, and I think the visualization looks good. My only comment is about how we present it.

It seems to me that the calibrations should be treated as an alternative to the score ranges for normal and abnormal phenotypes with their OddsPath values. When the user is configuring the score histogram, the choice of calibrations vs score ranges should be independent of the choice of histogram data. So it seems to me that the tabs across the top of the histogram, which govern the histogram data, should still be "Overall Distribution," "Clinical View," and "Custom." I don't think we ought to have "Calibrations" here.

Instead, I'd suggest just having something like what in this version is called "Shader Options," but maybe it should be called "Ranges" or something like that. There could be three options, depending what's available for a given score set:
- Score ranges (meaning the score ranges provided by the investigator)
- IGVF pillar project calibrations
- None
These might be supplemented by other calibrations in the future.

jstone-dev:
I also wanted to ask about the utility of the new `/score-sets/{urn}/calibration/data` endpoint. It seems like we're not using this yet and will just populate calibrations behind the scenes when they're available. I don't think it's a problem to have that endpoint anyway, but just wanted to make sure I understood its use (or non-use) correctly.

jstone-dev:
The latest changes look good to me. At some point we might need to revisit the default shader; if we had calibrations but no investigator-supplied ranges, we might want to default to the calibrations instead of no shader. But this seems unlikely to come up in practice.

PR #334, opened 2025-01-24
Updated MaveDB Citation

bencap:
Screenshot 2025-01-24 at 10 54 35 AM

### DIFF
--- a/src/components/screens/HomeScreen.vue
+++ b/src/components/screens/HomeScreen.vue
@@ -103,13 +103,13 @@
             </p>
             <blockquote>
                 Rubin, A.F., Stone, J., Bianchi, A.H. <i>et al.</i> MaveDB 2024: a curated community database with over seven million variant effects
-                from multiplexed functional assays. <i>Genome Biol</i> <b>26</b>, 13 (2025). <a href=https://doi.org/10.1186/s13059-025-03476-y>doi.org/10.1186/s13059-025-03476-y</a>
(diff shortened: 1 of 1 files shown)


src/components/screens/HomeScreen.vue
```diff
@@ -109,6 +94,26 @@
...
+            </p>
+            <blockquote>
+                Rubin, A.F., Stone, J., Bianchi, A.H. <i>et al.</i> MaveDB 2024: a curated community database with over seven million variant effects
+                from multiplexed functional assays. <i>Genome Biol</i> <b>26</b>, 13 (2025). <a href=https://doi.org/10.1186/s13059-025-03476-y>doi.org/10.1186/s13059-025-03476-y</a>
+            </blockquote>
+            Previous MaveDB publications are listed below:
+            <blockquote>
+              Daniel Esposito, Jochen Weile, Jay Shendure, Lea M Starita, Anthony T Papenfuss, Frederick P Roth, Douglas M Fowler, Alan F Rubin.
+              MaveDB: an open-source platform to distribute and interpret data from multiplexed assays of variant effect.
+              <i>Genome Biol</i> <b>20</b>, 223 (2019). <a href="https://doi.org/10.1186/s13059-019-1845-6">doi:10.1186/s13059-019-1845-6</a>
```
jstone-dev, line 112: The only thing I notice is that the DOI formats differ, with `doi.org/` and `doi:`. Maybe we can change one to be consistent. Looks like biomed gives the full URL as part of the citation format (`https://doi.org/10.1186/s13059-025-03476-y`), though it's probably okay to just start with `doi.org`.

bencap: Good point, agree they should be consistent. Pushed up a change for it.

Issue #213, opened 2024-06-03
Display partial publication details in input fields.


jstone-dev:
The "Publication identifiers" input field shows publication details in a drop-down menu, but after selecting a publication, only the identifier is visible. Showing a small portion of the title and/or the first author and year might be better for users.

jstone-uw:
Here's what the input looks like today while choosing and after choosing a publication:



![Image](https://github.com/VariantEffect/mavedb-ui/assets/74574922/9b1b5340-9366-49ac-af6e-fac6e2407988)


![Image](https://github.com/VariantEffect/mavedb-ui/assets/74574922/ea10af0b-8a65-4845-b709-0469b499c1c1)

bencap:
Perhaps show truncated title + pubmed / biorxiv / medrxiv identifier

PR #351, opened 2025-02-25
Solve #213 Display partial publication details in input fields.

EstelleDa:
This is an example effect. If the title is longer than 50 length, it'll be truncated. 
Screenshot 2025-02-25 at 7 45 10 pm


bencap (approved):
Thanks Estelle, looks good. I'm going to go ahead and merge this so we can include it in this release.

PR #352, opened 2025-03-03
Support for Base Editor Data

bencap:
Adds base editor support during creation and editing of score sets. Adds a note to the score set viewer under the targets when a score set represents base editor data.


bencap:
Supports https://github.com/VariantEffect/mavedb-api/pull/395

PR #354, opened 2025-03-10
Debug add a target gene creates an empty score ranges

EstelleDa:


src/components/screens/ScoreSetEditor.vue
```diff
@@ -278,6 +278,7 @@
               </template>
             </Card>
             <div v-if="itemStatus == 'NotLoaded' || this.item.private">
+              {{ scoreRanges }}
```
bencap, line 281: I think you missed removing this from debugging.
```suggestion
```

bencap (commented):
Thanks for looking at this Estelle, besides the one comment looks good to me.

src/components/screens/ScoreSetEditor.vue
EstelleDa: hahaha~~ Yes, you're correct!

PR #356, opened 2025-03-12
A solution to solve score set or experiment's title is too long problem.

EstelleDa:


bencap (commented):
To me I think it would make the most sense if the title wraps to multiple lines and then the buttons show up next to the last line of the title rather than having a scrollable title. I think we want people to be able to see the whole thing.

EstelleDa:
@bencap Sorry the for late reply. I agree showing the whole title is better. Otherwise, users may not notice the scroll button. Currently, the title will take all of the space if there isn't any button or collection.

src/assets/layout.css
```diff
@@ -15,22 +15,29 @@
...
 }
 
 .mave-screen-title-controls {
-  flex: 0 0 auto;
+  flex: 0 1 auto; /* Only take space if present */
+  align-self: flex-end;
+  display: flex;
 }
 
 .mave-screen-title-controls .p-button {
```
bencap, line 43: In this element, I like adding a line `width: max-content`. That makes the buttons maintain the full width of their text


Screenshot 2025-03-20 at 11 05 49 AM

bencap (commented):
Thanks for looking at this Estelle, the wrap looks much better to me. I left one minor comment about avoiding the buttons shrinking into each other.

I was wondering if as part of this you might also be able to move the `.mave-collection-badges` CSS element into the `layout.css` file. It seems to me like it would make sense as a shared element just like the other title bar elements.

I think it also makes sense to add a `margin-right: 5px` to the `.mave-collection-badges` just like we have for the buttons. That way, the collection badges have a margin against the buttons to their right.

EstelleDa:
My pleasure! :D

PR #358, opened 2025-03-17
Statistics Dashboard

bencap:
Adds a statistics dashboard screen. In support of this screen, a new time series line chart component has also been created. This component displays a new bar-line chart visualization, which takes an x, y1, and y2 object. This component displays the y1 elements as bars and the y2 elements on a line against a time series x axis.


jstone-dev:
I guess I did want to ask about an embeddable version of this for use on the AVE site, presumably without the navigation features. Do you think that will be necessary, and is the idea to try embedding the whole page first?

bencap:
> I guess I did want to ask about an embeddable version of this for use on the AVE site, presumably without the navigation features. Do you think that will be necessary, and is the idea to try embedding the whole page first?

Yeah, I think that's smart. The most recent commit adds conditional rendering to the default layout. If we want to suppress the toolbar we can just pass `withToolbar=False` to the `DefaultLayout` component

PR #362, opened 2025-03-19
Controlled keywords - gene ontology

EstelleDa:


EstelleDa:
Hi @afrubin @bencap , this is the draft codes of Gene Ontology. Feel free to review it when I get the final keyword list and may modify it. The accession will be presented together if the keyword has an accession as the below example.
image

PR #365, opened 2025-03-26
Updates for unspecified score range classifications

bencap:
Adds support for unspecified score range classifications. As part of this change, a new string formatter `beautifyString` was also added that converts underscores to spaces and makes the string title case. This better supports score ranges, which are stored in the database without spaces and in lower case, by making them appear in a more readable format.

Supports https://github.com/VariantEffect/mavedb-api/pull/414.


jstone-dev (approved):
Looks good to me.

I notice that the chart title and score range shading may overlap. I don't think this is new here:

Screenshot 2025-03-27 at 12 04 45 PM

PR #381, opened 2025-04-15
Modify histogram

EstelleDa:
1. Modify the histogram title position. 
2. Add `mave-histogram-controls` style codes. I don't think it's a typo so added this part. I only found `mave-histogram-control` in `<style scoped>`, but I think it's different from the `mave-histogram-controls`.
3. Align radio buttons to the same position. In the layout, the radio buttons, No Shader and Score Ranges, are not on the same line. 
4. Original codes `<div v-if="hasTabBar && showShaders" class="mave-histogram-controls">` didn't show the shader options because `hasTabBar` is `false` value while `showShaders` is `true`. And the histogram title is null if `hasTabBar` is `true`. Hence I think here is `!hasTabBar`. Or shall we either show the histogram title or the shader options?


bencap (commented):
Thanks for the changes Estelle, I like all the alignments better.

For the title thing though, I do think it makes the most sense to have the title still displayed by the histogram itself. For this particular bug, I think that means making changes to the title positioning within the histogram.ts file. I was playing around on my local instance and changing the y-attribute to `attr('y', -5)` looked ok, but I'm not sure if that will mess up the margins if those get changed.

https://github.com/VariantEffect/mavedb-ui/blob/ad33b6d43390be873dd22563d57e43362567099b/src/lib/histogram.ts#L711-L719

EstelleDa:
Thanks Ben! I tried your suggestions and it also works well on my side. I think modifying the title position within the histogram is a good way. Or how about setting the y value in another way? I put it in my new committed codes. Let me know what you think.

src/components/ScoreSetHistogram.vue
```diff
@@ -1,17 +1,17 @@
 <template>
   <TabMenu class="mave-histogram-viz-select" v-if="hasTabBar" v-model:activeIndex="activeViz" :model="vizOptions" />
-  <div v-if="hasTabBar && showShaders" class="mave-histogram-controls">
+  <div v-if="!hasTabBar && showShaders" class="mave-histogram-controls">
```
bencap, line 3: I believe this should remain as `hasTabBar`, otherwise we might show ranges when clinical features are disabled. I think the weirdness with this probably stems from line 449. I'm not sure why we would not want to display the title if the tab bar exists. So I think that line can probably change to `.title('Distribution of Functional Scores')`.

src/lib/histogram.ts
```diff
@@ -714,7 +714,7 @@ export default function makeHistogram(): Histogram {
             .join('text')
             .attr('class', 'histogram-title')
             .attr('x', width / 2 )
-            .attr('y', 4)
+            .attr('y', -margins.top / 2 + 5)
```
bencap, line 717: Yup, I like this much better. Perhaps we could even do `            .attr('y', -margins.top / 2 + (margins.top / 4))`? That way, the title is always 3/4 of the way into the margin and it isn't hardcoded to 5px even if the margin is like 200px (not sure why anyone would do that, but at least we'd handle it gracefully).

src/components/ScoreSetHistogram.vue
EstelleDa: I agree. I'll set it back to `hasTabBar` and modify the title part. Maybe it was a typo?

src/lib/histogram.ts
EstelleDa: Yup, this is better.

bencap (approved):

PR #389, opened 2025-05-07
Modify the tooltip position in heatmap

EstelleDa:
The tooltip won't be cut off.


bencap (approved):
Thanks Estelle! Let's merge this into the 2025.2.0 release since that will be the next thing to go out.

EstelleDa:
Sure!

PR #392, opened 2025-05-24
Remove hardcoded odds path values, use score range property to display them

bencap:
Removes hard coded odds path values from clinical alpha features.

Adds functionality to the wizard that allows the user to optionally submit odds path data with their score range. For now, evidence strengths are limited to PS3 and BS3 strong. The optional source for the score range data must be a publication associated with the score set.


jstone-dev (commented):
This looks good to me aside from the changes I mentioned in the API PR.

The "odds path source" line in the hard-coded mockup was never really requested; I just put it there to be clear that this was a precomputed value from a paper. I do think keeping it (optionally) as you've done is a nice feature.

src/components/OddsPathTable.vue
```diff
@@ -6,19 +6,19 @@
...
-                    <span v-if="range.oddsPath">{{ range.oddsPath.ratio }}</span>
+                <td v-for="range in abnormalRanges.concat(normalRanges)" :key="range.label">
+                    <span v-if="range.oddsPath?.ratio">{{ range.oddsPath.ratio }}</span>
                     <span v-else>Not Provided</span>
                 </td>
             </tr>
             <tr>
-                <td v-for="range in abnormalRanges.concat(normalRanges)" :class="`mave-evidence-code-${range.oddsPath?.evidence}`">
-                    <span v-if="range.oddsPath">{{ range.oddsPath.evidence }}</span>
+                <td v-for="range in abnormalRanges.concat(normalRanges)" :class="`mave-evidence-code-${range.oddsPath?.evidence}`" :key="range.label">
```
jstone-dev, line 20: I guess this will lead to a class name of `mave-evidence-code-` if there is no evidence code. That's probably fine, but we could do something to clean it up.

jstone-dev (approved):

src/components/OddsPathTable.vue
bencap: I think that because we have Pydantic set up to not return Nonetype properties, if we have an empty evidence strength value we will get `mave-evidence-code-undefined`, which I added a class for.

PR #400, opened 2025-05-29
Add confirmation dialog to score set publication

bencap:


src/components/screens/ScoreSetView.vue
```diff
@@ -675,34 +675,50 @@ export default {
...
+                path = path.slice(1)
+              }
+              path = path.join('.')
+              formValidationErrors[path] = error.msg
+            }
           }
-          path = path.join('.')
-          formValidationErrors[path] = error.msg
+        },
+        reject: () => {
```
jstone-dev, line 717: The reject option isn't required, but keep it if you think it makes the code clearer.

jstone-dev (approved):

PR #403, opened 2025-06-12
Drop empty hgvs columns from data tables.

EstelleDa:


src/components/screens/ScoreSetView.vue
```diff
@@ -781,24 +776,35 @@ export default {
...
+          const header = lines[0].split(',').map(h => h.trim())
+          const columnsToShow = header.filter(col => col !== 'accession')  // drop 'accession'
+
+          this.scoreColumns = columnsToShow
+
+          const fixedColumns = ['hgvs_nt', 'hgvs_splice', 'hgvs_pro']
+          this.sliceNumInDataTable = fixedColumns.filter(col => columnsToShow.includes(col)).length
+
+          const parsed = parseScoresOrCounts(response.data)
+          this.scoresTable = this.item.numVariants <= 10 ? parsed : parsed.slice(0, 10)
```
bencap, line 791: This might be a little crude, but since the list returned by the `parseScoreOrCounts` function derived from the CSV list, the first element of the list should have every column we need to display. This would bypass the need to deal with the CSV string directly, which would make this a little more readable, in my opinion. The code below would also apply to the counts.

```javascript
          const fixedColumns = ['hgvs_nt', 'hgvs_splice', 'hgvs_pro']

          this.scoresTable = parseScoresOrCounts(response.data)
          this.scoreColumns = Object.keys(this.scoresTable[0]).filter(col => col !== 'accession')  // drop 'accession'
          this.sliceNumInDataTable = fixedColumns.filter(col => this.scoreColumns.includes(col)).length

          this.scoresTable = this.item.numVariants <= 10 ? this.scoresTable : this.scoresTable.slice(0, 10)
```

PR #417, opened 2025-06-27
Davereinhart/protein visualization

davereinhart:
Add protein visualization component using PDBe-molstar and display alongside scoreset heatmap with interaction between the two to highlight residues on protein model and corresponding heatmap columns. Currently relies on UniProt ID provided by user on upload, but will eventually use UniProt ID determined by mapper function.


src/components/ProteinStructureView.vue
```diff
@@ -62,39 +89,81 @@ export default {
...
+          color: null,
+          focus: true
+        }))
+        this.viewerInstance.visual.select({data:[...this.selectionDataWithSelectedColorBy, ...selectedRanges]})
+        this.viewerInstance.visual.highlight({
+            data: selectedRanges,
+        })
+      },
+      deep: true,
+    },
```
bencap, line 158: Given that `this.viewerInstance` begins the lifecycle of this component as null, might it be worthwhile to guard its accession with `if (this.viewerInstance) { ... }`?

src/components/ProteinStructureView.vue
```diff
@@ -158,50 +202,59 @@ export default {
...
-        //
+
+        document.addEventListener('PDB.molstar.click', (e) => {
+          this.$emit('clickedResidue', e.eventData)
+        })
+        document.addEventListener('PDB.molstar.mouseover', (e) => {
+          this.$emit('hoveredOverResidue', e.eventData)
+        })
+        this.viewerInstance = viewerInstance
       }
```
bencap, line 245: We should probably clean up these event listeners in an [unmounted](https://vuejs.org/api/options-lifecycle#unmounted) hook. It seems like if they aren't explicitly cleaned up they'll persist even after the component has been unmounted.

src/components/ProteinStructureView.vue
```diff
@@ -62,39 +89,81 @@ export default {
...
+        if (this.selectedAlphaFold) {
+          newSelectedAlphaFold = this.alphaFoldData.find((x) => x.id == newSelectedAlphaFold.id)
         }
-        if (!this.selectedPdb && this.pdbs.length > 0) {
-          this.selectedPdb = this.pdbs[0]
+        if (!this.selectedAlphaFold && this.alphaFoldData.length > 0) {
+          this.selectedAlphaFold = this.alphaFoldData[0]
         }
       }
     },
```
bencap, line 169: Might not be understanding this fully, but could this be more simply written as the following?

```suggestion
     handler: function() {
        if (this.selectedAlphaFold) {
          this.selectedAlphaFold = this.alphaFoldData.find((x) => x.id == this.selectedAlphaFold.id)
       }
        if (!this.selectedAlphaFold && this.alphaFoldData.length > 0) {
          this.selectedAlphaFold = this.alphaFoldData[0]
       }
     }
   },
```

It seemed like we didn't need that temporary newSelectedAlphaFold variable / it wasn't doing anything.

src/components/screens/ScoreSetView.vue
```diff
@@ -108,14 +108,16 @@
             ref="histogram"
           />
         </div>
-        <div v-if="showHeatmap" class="mave-score-set-heatmap-pane">
+        <div v-if="showHeatmap && !isScoreSetVisualizerVisible" class="mave-score-set-heatmap-pane">
           <ScoreSetHeatmap
             :scoreSet="item"
             :scores="scores"
             :externalSelection="variantToVisualize"
+            :showProteinStructureModal="uniprotId!=null"
```
bencap, line 116: Since for now we'll only want to show the model opening option when the clinical preview settings are enabled, let's add another computed element:

```
    protein-structure-viewer-enabled: function() {
      return uniprotId != null && this.config.CLINICAL_FEATURES_ENABLED
    },
```

If it's not already, you may also need to import the config object and add it's ref to the data:
```
import config from '@/config'
...

data: function() {
    ...

    return {
      config: config,
      ...
    }
  },

**Review by bencap — COMMENTED** [review]
Thanks for working on this Dave, going to be a really good feature to get out! Left some comments with questions and a couple things to address before merging, but they should be pretty straightforward.

File: src/components/screens/ScoreSetView.vue (hunk 4 above)
Comment (davereinhart) [reply src/components/screens/ScoreSetView.vue:116]: Ok, Makse sense. I also realized this was a non-existent prop, what I actually wanted was `showProteinStructureButton`

For now, since this calc isn't likely to get more complex, I just put it directly on the prop:
```
:showProteinStructureButton="uniprotId!=null && config.CLINICAL_FEATURES_ENABLED"
```

src/components/ProteinStructureView.vue
davereinhart: Yes, they were persisting! Good catch. Removing them on `beforeUnmount` now.

PR #418, opened 2025-07-02
Fix the shader toggle doesn't work in histogram

EstelleDa:


bencap:
Thanks! There's one more spot where you'll need to change it to `notSet` on line 143 :)

PR #420, opened 2025-07-02
Adding fixed y-axis tick labels and legend to remain visible when scrolling heatmap horizontally

davereinhart:
Updates ScoreSetHeatmap component to include svg element on y-axis with absolute position, to remain visible when scrolling horizontally. The newly added svg element contains the legend and y-axis tick labels that are also present in the main svg element, and has the same positioning on initial load so that it is covering the original content.

The chart export function was also updated to exclude this element by adding "exclude-from-export" as a class name.

Also cleans up some unused class names and SVG child elements from the heatmap and ScoreSetHeatmap component.

### DIFF
--- a/src/lib/heatmap.ts
+++ b/src/lib/heatmap.ts
@@ -595,6 +595,9 @@ export default function makeHeatmap(): Heatmap {
         mainGroup.append('g')
           .attr('class', 'heatmap-nodes')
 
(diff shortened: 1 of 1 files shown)


bencap (approved):
lgtm, thanks for the work on this.

PR #428, opened 2025-07-25
Add a wizard completion page after creating a score set

EstelleDa:


EstelleDa:
@bencap Thanks for the feedback! They're quite helpful. I added a link of creating a new experiment to it. It looks a little redundant, but without it seems missing something to users so I put it there first. 
I also struggled whether need to jump to this page after editing a score set before. I think you're correct. They have seen this page already so it's not necessary to show it again.

bencap (approved):
Looks great!

PR #443, opened 2025-08-27
Rename tax_id to code in Taxonomy

EstelleDa:
The issue is https://github.com/VariantEffect/mavedb-api/issues/495
Back end PR is https://github.com/VariantEffect/mavedb-api/pull/500


bencap (changes requested):
There's one other file, `/src/schema/openapi.d.ts/components` that includes some references to taxId which we should change to `code`. Otherwise, looks good.

PR #525, opened 2025-10-20
ProteinStructureView: dynamic AlphaFold model file version

davereinhart:
Retrieves AlphaFold model version from API to be included in CIF file URL, rather than hardcoding, to prevent 404s when old file versions are purged by AlphaFold.

Goes with [#VariantEffect/mavedb-api/davereinhart/alphafold-files-version-proxy](https://github.com/VariantEffect/mavedb-api/pull/547/commits/590844e34c9b38527523ecca98a5df7c10fabc52)

### DIFF
--- a/src/components/ProteinStructureView.vue
+++ b/src/components/ProteinStructureView.vue
@@ -245,19 +245,32 @@ export default {
 
     render: async function () {
       if (this.selectedAlphaFold) {
-        let alphafoldFilesVersion
+        let alphafoldCifUrl
         try {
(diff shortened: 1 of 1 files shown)


davereinhart:
> Not explicitly necessary, but might it be worthwhile to add a toast notification or a page that says the alphafold file couldn't be loaded in cases where the request fails so that the page isn't just hanging?

Added checks and toast notifications on both related requests (AlphaFold version # and cif file)

PR #540, opened 2025-10-29
Include score and counts column metadata fields on ScoreSetCreator and ScoreSetEditor components

davereinhart:
Updates to correspond with API changes to include score and count columns metadata as inputs on ScoreSetCreator and ScoreSetEditor, and transition from PUT to PATCH endpoint for ScoreSetEditor.  With this change, user can upload JSON file for score and/or count column metadata which is parsed to JSON client side and submitted with score and count CSV files as part of multipart form data payload.

Goes with https://github.com/VariantEffect/mavedb-api/pull/546

### DIFF
--- a/src/components/screens/ScoreSetCreator.vue
+++ b/src/components/screens/ScoreSetCreator.vue
@@ -42,7 +42,12 @@
                         >.
                       </div>
                     </div>
-                    <div class="mavedb-wizard-content">Experiment title: {{ experiment.title }}</div>
+                    <div class="mavedb-wizard-content">
+                      Experiment title:
+                      <router-link :to="{name: 'experiment', params: {urn: experimentUrn}}">
+                        {{experiment.title}}
+                      </router-link>
+                    </div>
                   </div>
                 </div>
                 <div v-else>


src/components/screens/ScoreSetCreator.vue
```diff
@@ -1796,24 +1759,54 @@
...
-                      validationErrors.scoresFile
-                    }}</span>
+                    <span v-if="validationErrors.scoresFile" class="mave-field-error">{{validationErrors.scoresFile}}</span>
+                  </div>
+                </div>
+                <div class="mavedb-wizard-row">
+                  <div class="mavedb-wizard-help">
+                    <label :for="scopedId('input-scoreColumnsMetadataFile')">Load a scores column metadata file</label>
+                    <div class="mavedb-help-small">
+                      This file is optional, but recommended.
```
bencap: I guess I wonder if we might include some additional details here about the required structure of the metadata file. For instance: 
> This file is optional, but recommended. If provided, it should be a JSON file containing a single object. The keys of that object should be limited to columns of the score data, while values should include a string `description` and an optional string `details`. 

I'm not really in love with that description and the answer is probably that it should be added to the documentation, but I think we realistically haven't updated it in so long that bringing the docs up to date is a totally separate issue.

src/components/screens/ScoreSetCreator.vue
```diff
@@ -2881,31 +2896,57 @@ export default {
...
               try {
                 this.extraMetadata = JSON.parse(text)
-                if (!_.isObject(this.extraMetadata) || _.isArray(this.extraMetadata)) {
-                  this.clientSideValidationErrors.extraMetadata =
-                    'Extra metadata must be a JSON object (not an array or simple value).'
-                } else {
-                  this.clientSideValidationErrors.extraMetadata = null
-                }
-              } catch (e) {
+                this.validateExtraMetadata(this.extraMetadata, 'extraMetadata')
```
bencap: Function doesn't exist, did you mean `validateJsonObject`?

src/components/screens/ScoreSetCreator.vue
```diff
@@ -1796,24 +1759,54 @@
...
+                  </div>
+                  <div class="mavedb-wizard-content">
+                    <span class="p-float-label">
+                      <FileUpload
+                        :id="scopedId('input-scoreColumnsMetadataFile')"
+                        ref="scoreColumnsMetadataFileUpload"
+                        accept="application/json"
+                        :auto="false"
+                        choose-label="Scores column metadata file"
+                        :class="inputClasses.countsFile || ''"
```
bencap:
```suggestion
                        :class="inputClasses.scoresFile || ''"
```

src/components/screens/ScoreSetCreator.vue
```diff
@@ -1833,6 +1826,38 @@
...
+                </div>
+                <div class="mavedb-wizard-content">
+                  <span class="p-float-label">
+                    <FileUpload
+                      :id="scopedId('input-countColumnsMetadataFile')"
+                      ref="countColumnsMetadataFileUpload"
+                      accept="application/json"
+                      :auto="false"
+                      choose-label="Counts column metadata file"
+                      :class="inputClasses.countsFile || ''"
```
bencap:
```suggestion
                      :class="inputClasses.countsMetadataFile || ''"
```

src/components/screens/ScoreSetCreator.vue
```diff
@@ -2881,31 +2896,57 @@ export default {
...
+            {
+              try {
+                this.countColumnsMetadata = JSON.parse(text)
+                this.validateJsonObject(this.countColumnsMetadata, 'countColumnsMetadata')
+              } catch {
+                this.countColumnsMetadata = null
+                this.clientSideValidationErrors.countColumnsMetadata = 'The file did not contain valid JSON text.'
+                console.log('Count columns metadata file did not contain valid JSON text.')
+              }
+            }
```
bencap: Need a `break` here or setting the count metadata will also invoke the cases below.

bencap (commented):
Thanks for the changes and the housekeeping @davereinhart, I had no idea the creator file had so much obsolete duplication from the editor.

Given that we expect the metadata file to evolve, I don't think it's essential we render metadata in the UI as part of this task, but we should add another issue for it and start thinking about what that might look like. Another issue for adding a download button would be good as well in the interim and is probably all we need for a little while at least.

There was one blocking bug I found in the case statement that was causing the score metadata to be set as the same thing as the count metadata if the count metadata was set after the score metadata, but it should be a trivial fix. After that is all set this looks good.

src/components/screens/ScoreSetCreator.vue
```diff
@@ -109,13 +100,7 @@
                       </div>
                     </div>
                     <div class="mavedb-wizard-content">
-                      <div v-if="itemStatus != 'NotLoaded' && item.experiment">
-                        Experiment:
-                        <router-link :to="{name: 'experiment', params: {urn: item.experiment.urn}}">{{
-                          item.experiment.title
-                        }}</router-link>
-                      </div>
-                      <div v-else style="position: relative">
```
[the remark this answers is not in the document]
davereinhart: I thought this element was only relevant in the context of `ScoreSetEditor` component, and the `ScoreSetCreator` should always display the experiment dropdown, which is why I removed it here (assumed it was an artifact of when these two were one component).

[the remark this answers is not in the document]
davereinhart: Thanks for pointing out that experimentUrn can be set via ScoreSetCreator props. Readded the link to display when experiment is set.

bencap (approved):

PR #574, opened 2025-11-18
Replace Alphafold URL HEAD request with check of PDBe Molstar viewer …

davereinhart:
…strucureRefMap

To catch and display UI errors related to invalid or otherwise unavailable AlphaFold CIF file URLs that are being used in PDBe Molstar viewer of ProteinStructureView component, the previous solution of making a HEAD request was not behaving consistently. In some cases, CORS errors were occuring, but no obvious pattern was observed.

To address this, the HEAD request is being replaced by a check of the size of the PDBe molstar viewer instance's structureRefMap property. This value should be greater than 0 if the protein structure loaded successfully, with an error message displayed otherwise.

### DIFF
--- a/src/components/ProteinStructureView.vue
+++ b/src/components/ProteinStructureView.vue
@@ -255,18 +255,6 @@ export default {
           return
         }
 
-        // verify that the alphafoldCifUrl exists by verifying 200 or 304 status
-        try {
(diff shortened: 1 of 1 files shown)


bencap (approved):
Clever! Thanks for the improvement.

PR #576, opened 2025-11-19
Add upload your data section on Home page

EstelleDa:


src/components/screens/CollectionsView.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <DefaultLayout>
+  <DefaultLayout v-if="!userIsAuthenticated">
+    <p>You may <a href="#" @click.prevent="signInCollection">sign in</a> to view or add your collections.</p>
+  </DefaultLayout>
+  <DefaultLayout v-else>
```
bencap: I like this as a feature, but I think we can reduce all the code duplication here by implementing it at the level of the `DefaultLayout`. What I have in mind is:
- Adding an optional prop `requireAuth` to the `DefaultLayout.vue` component
- Import the auth module to `DefaultLayout.vue` as you do in these files to gain access to the `userIsAuthenticated` and `signIn` methods.
- Guard the slot with a `v-if="userIsAuthenticated"`:
```
      <div :class="mainClasses">
        <template v-if="userIsAuthenticated">
            <slot />
        </template>
        <template v-else>
            <div>information about needing to sign in to access this component</div>
        </template>
        <Footer v-if="footer == 'flow'" />
      </div>
```

We admittedly lose out on some information about the route being in the help text, but I think the user should understand what they are meant to see if they are on the page anyway.

bencap: Totally agree, it was just that if we implement it at the DefaultLayout level instead of the individual component level we don't have to duplicate the code and it's much easier to add the auth requirement to future components that might need it.

src/components/screens/HomeScreen.vue
```diff
@@ -130,15 +141,46 @@
...
+        signIn()
+      }
+    }
+    return { 
+      addScoreSet, 
+      addExperiment,
+      userIsAuthenticated
+    }
+  },
+
```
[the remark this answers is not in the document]
EstelleDa: Thanks Ben! I'll remove the unnecessary part, but we need the `sessionStorage.setItem("postLoginRedirect", "createExperiment")`. Otherwise, the page will stay on the Home page after clicking the Add Experiment/Score Set buttons.

src/components/layout/DefaultLayout.vue
```diff
@@ -58,6 +71,15 @@ export default defineComponent({
     }
   },
 
+  setup: () => {
+    const {signIn, userIsAuthenticated} = useAuth()
+
```
bencap: We could actually add something like the following here:
```typescript
    const signInWithRedirect = () => {
      const currentFullPath = window.location.pathname + window.location.search + window.location.hash
      localStorage.setItem('redirectAfterLogin', currentFullPath)
      signIn()
    }
```

Then if we use `signInWithRedirect` above in the login hint, we will redirect users back to the current page.

PR #590, opened 2025-12-04
Add molecular mechanism assessed controlled keywords to both Create Exp and Edit Exp pages.

EstelleDa:
UI PR of https://github.com/VariantEffect/mavedb-ui/issues/591


bencap (approved):
Looks good, thanks Estelle. I don't think that adding the `Other` entry to the database has any effect on these changes, but we should remember to do that when we release this.

PR #594, opened 2025-12-10
PrimeVue upgrade and addition of TailwindCSS 

davereinhart:
This branch upgrades PrimeVue from v3 to v4 and adds TailwindCSS package to simplify styling going forward.

This upgrade impacted a number of pages due to breaking changes related to the redesign of the Stepper component in PrimeVue 4, as well as differences in themes/CSS between the two versions. The TabView component is also marked as deprecated in PrimeVue 4 and is being replaced by the Tabs component.

Changes in this branch seek to maintain the look and feel of the current version of MaveDB, which required retaining the PrimeVue3 theme.css file. Use of PrimeVue3's CSS with PrimeVue4 isn't ideal, but it is used here to expedite the transition without drastically changing the current style of the site or reworking every component. An effort to fully transition to PrimeVue 4 theme and tailwind CSS classes, and phase out the use of the theme.css file is warranted, but can be tackled in a separate branch or more gradually as part of future releases.

### DIFF
--- a/src/assets/app.css
+++ b/src/assets/app.css
@@ -121,3 +121,11 @@ body .p-button {
   gap: 0 !important;
   margin-left: 1em !important;
 }
+
+.p-tab {
+  padding: 0.5em 1em 0.5em 1em !important;
+}
+
+.p-listbox-header {
+  padding: 0 !important;
+}
--- a/src/components/CalibrationEditor.vue
+++ b/src/components/CalibrationEditor.vue
@@ -229,6 +229,7 @@
                 :aria-labelledby="scopedId(`input-investigatorProvidedRangeLower-${rangeIdx}`)"
                 class="w-full"
                 :disabled="functionalRangeHelpers[rangeIdx].infiniteLower"
+                :max-fraction-digits="10"
               />
               <label :for="scopedId(`input-investigatorProvidedRangeLower-${rangeIdx}`)">{{
                 functionalRangeHelpers[rangeIdx].infiniteLower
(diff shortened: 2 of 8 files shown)


davereinhart:
Thanks for the detailed review @bencap! Some of these I should have noticed but I clearly didn't do enough testing. I had assumed that the components that were only marked as deprecated but were still available in PrimeVue 4 would still work, but that's obviously not the case.

I've swapped out most of those now with the exception of the Cards component. According to the migration documentation, those should be replaced by AutoComplete with multiple="true" but I wasn't able to get it working the same as before so instead I made changes that restored the PrimeVue 3 Cards functionality and style.

So there will still be some work to do to scrub the remaining deprecated components, the PrimeVue 3 theme, and custom CSS in favor of PrimeVue 4 and TailwindCSS. But this version should at least let us move forward with the upgrade and keep the existing functionality and style intact.

bencap (commented):
Awesome, thanks for the changes. Sorry, a couple more things as I am going through again:

- During score set creation, the float label obscures the text hint for the primary publication identifier. The experiment dropdown is working as intended.

Image

- During score set creation, I am unable to enter a decimals in the functional ranges. I think that there's a special prop on the primevue component you need to set to allow decimal values.

- During score set creation, the last file entry appears to be styled differently.

Image

- On the settings page, the `Welcome to MaveDB` text is smaller than on our current branch.

- The tab panels on the `Dashboard` and `Search` pages should have defaults so it's clearer what is shown.

Other than those items, I think this looks great! Should we open up some more targeted issues for the notes you commented with regard to scrubbing the deprecated components and styling the site with Tailwind?

bencap:
One other thing: The secondary references on the score set page appear misaligned. They look good on the experiment page though.
Screenshot 2025-12-16 at 3 33 09 PM

davereinhart:
Thanks again! Those changes have been made. I will add an issue for the additional work to scrub PrimeVue 3 components and styles completely, though in the case of Cards components there may not be an exact equivalent.

PR #603, opened 2025-12-23
chore: remove home link from toolbar

bencap:
This pull request removes the "Home" navigation button from the `Toolbar.vue` component. Some minor automatic styling changes are also included.

**Navigation updates:**
 - Removed the "Home" button from the toolbar navigation menu.


davereinhart (approved):
Rebased with latest 2025.6.0 plus added a commit to replace Toolbar inline styles with equivalent tailwind classes.

PR #607, opened 2026-01-06
Remove remaining PrimeVue 3 specific components and CSS, and improve score set and experiment validator error handling

davereinhart:
This branch completes the migration to PrimeVue 4 by replacing any remaining components from PrimeVue 3 that are marked as deprecated or have been renamed, to cover all of the items in the migration guide found here: https://primevue.org/guides/migration/v4

In addition, some improvements have been made on logic and display of validation errors for the `ScoreSetCreator`, `ScoreSetEditor` components, and in the separation of functionality for the `ExperimentCreator` and `ExperimentEditor` components.

Goes with https://github.com/VariantEffect/mavedb-api/pull/635

### DIFF
--- a/src/components/screens/ScoreSetCreator.vue
+++ b/src/components/screens/ScoreSetCreator.vue
@@ -2708,8 +2708,6 @@ export default {
             }
 
             if (_.isEqual(_.first(path), 'targetGenes') && path.length >= 2) {
-              // this.minTargetGeneStepWithError = Math.min(this.minTargetGeneStepWithError, path[1])
-
               // Map errors on indexed external gene identifiers to inputs named for the identifier's database.
               if (_.isEqual(_.slice(path, 2, 3), ['externalIdentifiers'])) {
                 const identifierIndex = path[3]
--- a/src/components/CalibrationEditor.vue
+++ b/src/components/CalibrationEditor.vue
@@ -960,12 +960,6 @@ export default {
       }
     },
 
-    // clearPublicationIdentifierSearch: function (refName: string) {
-    //   // This could change with a new Primevue version.
-    //   const input = this.$refs[refName]
-    //   input.$refs.focusInput.value = ''
-    // },
-
     searchPublicationIdentifiers: function (event: {query: string}) {
       const searchText = (event.query || '').trim()
       if (searchText.length > 0) {


src/components/screens/ScoreSetCreator.vue
```diff
@@ -2695,8 +2671,8 @@ export default {
               path = path.concat(customPath)
             }
 
-            if (_.isEqual(_.slice(path, 0, 1), ['targetGene'])) {
-              this.minTargetGeneStepWithError = Math.min(this.minTargetGeneStepWithError, path[1])
+            if (_.isEqual(_.first(path), 'targetGenes') && path.length >= 2) {
+              // this.minTargetGeneStepWithError = Math.min(this.minTargetGeneStepWithError, path[1])
```
bencap: Can we just remove this line if not being used?

bencap (commented):
Looks great overall, and thanks again for all the work on this Primevue upgrade.

Main blocking issue:
- The score column metadata and count column metadata file uploads remain disabled even after adding a score / count file.

A couple other small things I noticed, all minor:
- The tag text color isn't white anymore, which reduces contrast (see https://127.0.0.1:8081/score-sets/urn:mavedb:00000097-0-2?calibration=urn:mavedb:calibration-53755a6c-426b-42f4-b771-80283ac41aa2)
- OrcID elements for creators and modifiers, plus elements like the UCSC logo are on different lines from their associated text (see https://127.0.0.1:8081/score-sets/urn:mavedb:00000097-s-1).
- The 'Back' button color on the score set creator and experiment creator makes it look disabled.
- On the score set creator (and other places with the toggle switch) the text at right of all the toggle switches is no longer center aligned.
- I think the messages would look better with a small top margin (top of the score set information page, when selecting a non-default license, top of the page for score sets with mapping issues, etc.).
- I still see the `Users` item in the toolbar even though the component was removed.
- The 'Close' button color on the score calibration creator makes it look disabled.
- When editing a score set, the message for autopopulating fields still shows up but that probably isn't relevant outside of the score set creator (this might be a problem unrelated to your changes).

davereinhart:
Thanks for the feedback @bencap . I addressed your comments, so let me know if those parts look better now and if there's anything else.

PR #623, opened 2026-02-19
feat: migrate class variants to lazy endpoints and update class-view handling

bencap:
- consume new calibration variant endpoints via API helpers
- switch functional classification usage to id/variantCount model
- replace embedded variant counts in calibration table with variantCount
- lazily load class-based calibration variants in histogram view with loading state and cache
- fix class-view empty state by gating series on loaded class map and triggering refresh on load
- add TODO#622 note for future large-dataset performance optimizations

### DIFF
--- a/src/components/CalibrationTable.vue
+++ b/src/components/CalibrationTable.vue
@@ -144,9 +144,7 @@
                 }
               ]"
             >
-              <span>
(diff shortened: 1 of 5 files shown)


bencap:
UI support for https://github.com/VariantEffect/mavedb-api/pull/663

Issue #630, opened 2026-02-25
Protein structure view fails when AlphaFold version endpoint is unavailable


PlushZ:
Clicking "View protein structure" under the heatmap fails with the error:
`Failed to fetch AlphaFold version`

In ProteinStructureView.vue, rendering currently depends on ${config.apiBaseUrl}/alphafold-files/version. If that endpoint is missing or temporarily unavailable, the component exits and never attempts to load the structure.

Expected behavior:
Protein structure view should still attempt to render when version lookup fails.

Observed behavior:
The component returns early after version fetch failure, so users cannot access protein structure view.

Issue #512, opened 2025-10-13
Hardcoded AlphaFold Version #


bencap:
We currently have the AlphaFold version # hardcoded in our src: https://github.com/VariantEffect/mavedb-ui/blob/f82b6ffa16b05275983a4923260df0d9dc7efaf1/src/components/ProteinStructureView.vue#L250

This `/files` bucket is for the `latest-version` only though, so when the version updates on AlphaFold's end, we get 404 errors as our hardcoded version is evicted from the bucket.

We should find a more flexible way to handle this version number, ideally by always be fetching the latest version of CIF files from the latest-version bucket.

PR #631, opened 2026-02-25
fix alphafold version fetch failure

PlushZ:
Made AlphaFold model version resolution resilient in ProteinStructureView by adding a fallback to v6 when /alphafold-files/version is unavailable. This removes an early failure path that blocked protein structure rendering.
Solves the issue https://github.com/VariantEffect/mavedb-ui/issues/630
Can be related to https://github.com/VariantEffect/mavedb-ui/issues/512 and https://github.com/VariantEffect/mavedb-ui/pull/525

### DIFF
--- a/src/components/ProteinStructureView.vue
+++ b/src/components/ProteinStructureView.vue
@@ -246,12 +246,9 @@ export default {
       this.$emit('hoveredOverResidue', e.eventData)
     },
     fetchAlphaFoldCifUrl: async function () {
-      const qualifier = encodeURIComponent(this.selectedAlphaFold.id)
-      const response = await axios.get(`https://alphafold.ebi.ac.uk/api/prediction/${qualifier}`)
+      const response = await axios.get(`https://alphafold.ebi.ac.uk/api/prediction/${this.selectedAlphaFold.id}`)
       const predictionModels = _.isArray(response.data) ? response.data : [response.data]
-      const selectedModel =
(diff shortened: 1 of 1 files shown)


src/components/ProteinStructureView.vue
```diff
@@ -246,16 +245,26 @@ export default {
     hoveredOverResidue: function (e) {
       this.$emit('hoveredOverResidue', e.eventData)
     },
+    fetchAlphaFoldCifUrl: async function () {
+      const qualifier = encodeURIComponent(this.selectedAlphaFold.id)
+      const response = await axios.get(`https://alphafold.ebi.ac.uk/api/prediction/${qualifier}`)
+      const predictionModels = _.isArray(response.data) ? response.data : [response.data]
+      const selectedModel =
+        predictionModels.find((x) => _.isString(x?.cifUrl) && _.isString(x?.entryId) && x.entryId === `AF-${this.selectedAlphaFold.id}-F1`) ||
+        predictionModels.find((x) => _.isString(x?.cifUrl))
```
davereinhart: Here we only want to use the isoform specified by the uniprot ID, because we need the chosen isoform to match the mapped variants. So if there's not an exact match, we don't want to display the protein structure. So this can probably be simplified to: 
```
const selectedModel = predictionModels.find((x) => x.entryId === `AF-${this.selectedAlphaFold.id}-F1`)
```
OR
```
const selectedModel = predictionModels.find((x) => x.uniprotAccession === this.selectedAlphaFold.id)
```

davereinhart (changes requested):

PR #649, opened 2026-04-10
Add controlled keywords to score set search filter options.

EstelleDa:
Front end of https://github.com/VariantEffect/mavedb-api/issues/693.

### DIFF
--- a/src/components/common/MvSearchFilters.vue
+++ b/src/components/common/MvSearchFilters.vue
@@ -65,7 +65,7 @@
         @update:model-value="$emit('update:publicationDatabases', $event)"
       />
     </MvCollapsible>
-    <MvCollapsible :open="false" title="Keyword">
+    <MvCollapsible :open="false" title="Keywords">
       <div
         v-for="group in groupedControlledKeywordOptions"
         :key="group.key"
@@ -97,20 +97,6 @@ export default defineComponent({
 
   components: {MvCollapsible, MvSelectList},
 
-  computed: {
-    groupedControlledKeywordOptions(): { key: string; options: FilterOption[] }[] {
-      const groups: Record<string, FilterOption[]> = {}
-      for (const opt of this.controlledKeywordOptions) {
-        const key = opt.groupKey || 'Other'
-        if (!groups[key]) {
-          groups[key] = []
-        }
-        groups[key].push(opt)
-      }
-      return Object.entries(groups).map(([key, options]) => ({key, options}))
-    }
-  },
-
   props: {
     controlledKeywords: {type: Array as PropType<string[]>, default: () => []},
     targetNames: {type: Array as PropType<string[]>, default: () => []},
@@ -132,10 +118,35 @@ export default defineComponent({
     loading: {type: Boolean, default: false}
   },
 
+  emits: [
+    'update:controlledKeywords',
+    'update:targetNames',
+    'update:targetOrganismNames',
+    'update:targetTypes',
+    'update:targetAccession',
+    'update:publicationAuthors',
+    'update:publicationJournals',
+    'update:publicationDatabases'
+  ],
+
   data() {
     return {
       // per-group local selections
       groupSelections: {} as Record<string, string[]>
+    }
+  },
+
+  computed: {
+    groupedControlledKeywordOptions(): { key: string; options: FilterOption[] }[] {
+      const groups: Record<string, FilterOption[]> = {}
+      for (const opt of this.controlledKeywordOptions) {
+        const key = opt.groupKey || 'Other'
+        if (!groups[key]) {
+          groups[key] = []
+        }
+        groups[key].push(opt)
+      }
+      return Object.entries(groups).sort(([aKey], [bKey]) => aKey.localeCompare(bKey)).map(([key, options]) => ({key, options}))
     }
   },
 
@@ -157,17 +168,6 @@ export default defineComponent({
       }
     }
   },
-
-  emits: [
-    'update:controlledKeywords',
-    'update:targetNames',
-    'update:targetOrganismNames',
-    'update:targetTypes',
-    'update:targetAccession',
-    'update:publicationAuthors',
-    'update:publicationJournals',
-    'update:publicationDatabases'
-  ],
 
   methods: {
     onGroupChange(groupKey: string, values: string[]) {
--- a/src/components/common/MvSearchFilters.vue
+++ b/src/components/common/MvSearchFilters.vue
@@ -65,6 +65,15 @@
         @update:model-value="$emit('update:publicationDatabases', $event)"
       />
     </MvCollapsible>
+    <MvCollapsible :open="false" title="Keyword">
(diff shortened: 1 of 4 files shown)


src/components/common/MvSearchFilters.vue
```diff
@@ -65,6 +65,15 @@
         @update:model-value="$emit('update:publicationDatabases', $event)"
       />
     </MvCollapsible>
+    <MvCollapsible :open="false" title="Keyword">
```
bencap:
```suggestion
    <MvCollapsible :open="false" title="Keywords">
```

src/components/common/MvSearchFilters.vue
```diff
@@ -79,14 +97,30 @@ export default defineComponent({
...
+      const groups: Record<string, FilterOption[]> = {}
+      for (const opt of this.controlledKeywordOptions) {
+        const key = opt.groupKey || 'Other'
+        if (!groups[key]) {
+          groups[key] = []
+        }
+        groups[key].push(opt)
+      }
+      return Object.entries(groups).map(([key, options]) => ({key, options}))
+    }
```
bencap: I think we should consider ordering this list alphabetically. Right now, selecting a keyword can change the group order and it's a bit jarring from a user perspective.

src/components/common/MvSearchFilters.vue
```diff
@@ -98,14 +132,52 @@ export default defineComponent({
...
+            const key = opt.groupKey || 'Other'
+            if (!byGroup[key]) byGroup[key] = []
+            byGroup[key].push(opt.value)
+          }
+        }
+        this.groupSelections = byGroup
+      }
+    }
+  },
+
```
bencap, line 171: I believe this block is responsible for a bug where selecting any keyword `Other` also selects all the other `Other` keywords from each keyword group.

I think that the most precise way around many `Other` options existing is to make the a composite identifier. This would look like `key::value` or `MolecularMechanismAssessed::Other`. Then before we send to the API, we split on the `::` so we can send both the key and the value.

We'd need some additional changes, which I'll put in other comments.

src/components/screens/SearchView.vue
```diff
@@ -596,7 +612,12 @@ export default defineComponent({
           requestParams,
           controller.signal
         )
-
+        
+        this.controlledKeywordOptions = (data.controlledKeywords || []).map((option) => ({
+          value: option.value,      
+          badge: option.count,
+          groupKey: option.key
```
bencap: Here we could add a `title` entry which holds just `options.value`. Then `value` becomes `${option.key}::${option.value}` to disambiguate the keywords across groups.

src/lib/search.ts
```diff
@@ -74,7 +75,7 @@ export function buildSearchParams(filters: SearchFilters) {
     authors: filters.filterPublicationAuthors.length > 0 ? filters.filterPublicationAuthors : undefined,
     databases: filters.filterPublicationDatabases.length > 0 ? filters.filterPublicationDatabases : undefined,
     journals: filters.filterPublicationJournals.length > 0 ? filters.filterPublicationJournals : undefined,
-    keywords: filters.filterKeywords.length > 0 ? filters.filterKeywords : undefined
+    controlledKeywords: filters.filterControlledKeywords.length > 0 ? filters.filterControlledKeywords : undefined,
```
bencap: Before sending to the backend, we'd now need to split on `::`

src/components/screens/SearchView.vue
```diff
@@ -320,6 +326,7 @@ export default defineComponent({
           filters.push({key, value, label: labelFn ? labelFn(value) : value})
         }
       }
+      addFilters('controlledKeywords', this.filterControlledKeywords)
```
bencap: We'd need a custom `labelFn` here that strips the prefix before `::` so that we get plain values for display or we could pass through the `{key, value}` object and extract just `value`.

bencap (changes requested):
Thanks for working on this Estelle. The main comment I have is about the behavior when a user selects a keyword whose value is present across keyword groups, such as 'Other'. We'll need some additional backend changes to fully remedy it.

src/components/screens/SearchView.vue
```diff
@@ -320,6 +326,7 @@ export default defineComponent({
           filters.push({key, value, label: labelFn ? labelFn(value) : value})
         }
       }
+      addFilters('controlledKeywords', this.filterControlledKeywords, (v) => v.split('::', 2)[1] ?? v)
```
bencap: To keep with convention, this should be `filterControlledKeywords`.
```suggestion
      addFilters('filterControlledKeywords', this.filterControlledKeywords, (v) => v.split('::', 2)[1] ?? v)
```

src/components/screens/SearchView.vue
bencap, line 652: My note above also brought forward a bug in my initial implementation here, which is that the array manipulation we are doing means that Vue isn't properly reacting to changes in the filter arrays.

I was testing things on my local and this is what I would suggest implementing:
```suggestion
   removeFilter(key: string, value: string) {
      const setters: Record<string, (v: string[]) => void> = {
        filterControlledKeywords: (v) => (this.filterControlledKeywords = v),
        filterTargetNames: (v) => (this.filterTargetNames = v),
        filterTargetOrganismNames: (v) => (this.filterTargetOrganismNames = v),
        filterTargetTypes: (v) => (this.filterTargetTypes = v),
        filterTargetAccession: (v) => (this.filterTargetAccession = v),
        filterPublicationAuthors: (v) => (this.filterPublicationAuthors = v),
        filterPublicationDatabases: (v) => (this.filterPublicationDatabases = v),
        filterPublicationJournals: (v) => (this.filterPublicationJournals = v)
      }
      const getters: Record<string, string[]> = {
        filterControlledKeywords: this.filterControlledKeywords,
        filterTargetNames: this.filterTargetNames,
        filterTargetOrganismNames: this.filterTargetOrganismNames,
        filterTargetTypes: this.filterTargetTypes,
        filterTargetAccession: this.filterTargetAccession,
        filterPublicationAuthors: this.filterPublicationAuthors,
        filterPublicationDatabases: this.filterPublicationDatabases,
        filterPublicationJournals: this.filterPublicationJournals
      }
      const setter = setters[key]
      const arr = getters[key]
      if (!setter || !arr) return
      const filtered = arr.filter((v) => v !== value)
      setter(filtered)
   },
```

Using this setter/getter map means that the arrays get replaced, so Vue will properly react to changes in the filter lists and update the search.

bencap (commented):
Thanks for the changes Estelle. There was one more bug I noticed, but it was more just that the keyword changes made me notice a bug in the original code that wasn't happening without the unique structure of the keywords.

src/components/screens/SearchView.vue
EstelleDa: Thanks for finding this! I didn't notice it.

bencap (approved):
Looks great!

bencap (approved):

PR #670, opened 2026-05-06
Add a condition check to the collection section.

### LINKED ISSUE #643: Users without edit access on collections have UI elements which imply edit access
To reproduce:
- Navigate to a collection without either score sets or experiments which you can view but not edit
- Observe a empty state message which implies you can add data sets to the collection

On collection pages, we should ensure appropriate empty state messages for users with different permission levels.

EstelleDa:


bencap (changes requested):
Thanks for looking at this Estelle. Although this implementation does work, I think there's a simpler way using the props that already exist which minimizes the size of this change.

Callers pass `canUpdate` and `canAdd` to the collection items table, which means we should already know whether a user can add or update the collection without checking permission again.

Looking at the `MvEmptyState` component, it handles interactivity by checking if there is a `to` prop and a `action-label` prop: https://github.com/VariantEffect/mavedb-ui/blob/d2a88ee2421590efd8aef854aab85581acb22167/src/components/common/MvEmptyState.vue#L37-L39.

Since we don't have a `to` prop here, that means we can control the interactivity of the MvEmptyState component by controlling whether it gets a `action-label`. Since we have the `canAdd` prop, we can use that to conditionally inject an action label; line 32 would become ``:action-label="canUpdate ? `Add ${entityTypeLabel}s` : undefined"``.

Handling our interactivity like this, we only have to make one small addition to the component, we make use of the existing props and don't have to recompute permissions, and the displayed empty state is consistent with the design of the rest of the site.

EstelleDa:
> Thanks for looking at this Estelle. Although this implementation does work, I think there's a simpler way using the props that already exist which minimizes the size of this change.
> 
> Callers pass `canUpdate` and `canAdd` to the collection items table, which means we should already know whether a user can add or update the collection without checking permission again.
> 
> Looking at the `MvEmptyState` component, it handles interactivity by checking if there is a `to` prop and a `action-label` prop:
> 
> https://github.com/VariantEffect/mavedb-ui/blob/d2a88ee2421590efd8aef854aab85581acb22167/src/components/common/MvEmptyState.vue#L37-L39
> 
> .
> Since we don't have a `to` prop here, that means we can control the interactivity of the MvEmptyState component by controlling whether it gets a `action-label`. Since we have the `canAdd` prop, we can use that to conditionally inject an action label; line 32 would become `` :action-label="canUpdate ? `Add ${entityTypeLabel}s` : undefined" ``.
> 
> Handling our interactivity like this, we only have to make one small addition to the component, we make use of the existing props and don't have to recompute permissions, and the displayed empty state is consistent with the design of the rest of the site.

Thanks Ben! I think it's much better.

src/components/collection/CollectionItemsTable.vue
```diff
@@ -61,6 +62,10 @@ export default defineComponent({
       type: Array,
       required: true
     },
+    itemId: {
+      type: String,
+      required: true
+    },
```
bencap: Do we still need this itemId?

EstelleDa: Opps. Nope.

bencap (approved):

Issue #683, opened 2026-05-21
Refactor MaveMD search state and execution flow


bencap:
## Summary

The MaveMD variant search flow currently works, but the implementation has accumulated multiple sources of truth for search mode, route state, form state, and search execution. This makes small behavior changes harder to reason about and increases the risk of state leaks between standard search and Guided Search.

This issue proposes a focused cleanup to simplify the search path without changing user-facing behavior.

## Problem

The current implementation in `SearchVariantsScreen.vue` is state-heavy and relies on several side effects that are easy to misunderstand:

- Search mode is represented by both `defaultSearchVisible` and `guidedSearchVisible`, while the route also represents guided mode with `mode=guided`.
- `clearSearch()` clears local form/results state and also replaces the full query with `{}`, so callers that only want to clear local state can accidentally wipe route state.
- `fetchDefaultSearchResults()` both dispatches by search type and executes HGVS/dbSNP/ClinVar/ClinGen/VRS lookups. Guided Search constructs HGVS strings, but currently has to call through this dispatcher.
- `searchType` is ambient component state read at lookup time, which means callers must ensure it is correct before calling shared search helpers.
- The default search type and search modes are represented by repeated string literals such as `'hgvs'`, `'guided'`, and `'default'`.

These issues are orthogonal to the bug fix in #675, but they make future maintenance more fragile.

## Proposed behavior

Refactor the MaveMD search flow so each piece of state has a clearer responsibility while preserving current behavior:

1. Separate local search-state clearing from route clearing.
2. Introduce explicit constants for default search type and search modes.
3. Prefer a single source of truth for active search mode.
4. Consider extracting HGVS lookup execution from the broader default-search dispatcher so Guided Search can call HGVS lookup directly.

No user-facing behavior should change as part of this cleanup.

## Acceptance criteria

- Standard search still supports HGVS, ClinGen CAId, ClinVar ID, dbSNP rsID, and VRS digest searches with the same validation and result behavior as before.
- Guided Search still syncs its fields to route query params and remains deep-linkable via `mode=guided`.
- Switching between standard search and Guided Search does not leave stale route params or stale form values behind.
- Clearing a search still clears results and form state, but route-clearing behavior is explicit at the call site.
- Search mode cannot enter an impossible UI state where both standard and Guided Search are visible or both are hidden.
- Guided Search uses HGVS lookup behavior directly or through an explicitly named HGVS path, rather than depending on ambient `searchType` state.
[...]

PR #684, opened 2026-05-21
Guided Search UX Improvements

### LINKED ISSUE #675: Guided Search: Add Homepage Access, Fix Case Sensitivity UX, and Fix Search Type State Bug
## Summary

Three related improvements to the Guided Search feature in the MaveMD variant search (`/mavemd`):

1. The homepage hero section links to individual standard search methods (HGVS, ClinGen CAId, dbSNP rsID, ClinVar ID, VRS Digest) but provides no path to the Guided Search mode, making it hard to discover.
2. The gene symbol input does not visually indicate that it is case-insensitive, causing user confusion even though the input is silently uppercased before the API call.
3. When a user enters Guided Search after having selected a non-HGVS tab in standard search mode, the `searchType` state is not reset, causing the guided search to submit the constructed HGVS string against the wrong search type (e.g., dbSNP rsID), producing broken results.

---

### 1. No Homepage Entry Point for Guided Search

`HomeScreen.vue` renders a row of quick-links below the hero search bar pointing directly to `/mavemd?searchType=<type>` for each standard search type. There is no equivalent entry point for Guided Search (`/mavemd?mode=guided`). Users unfamiliar with HGVS notation — the primary target audience for Guided Search — are the least likely to discover it by exploring the MaveMD page on their own.

### 2. Case Sensitivity Not Communicated to User

In `SearchVariantsScreen.vue`, the gene symbol input is converted to uppercase via `geneSymbol?.toUpperCase()` immediately before the ClinGen API call. However, no placeholder text, helper label, or visual normalization (e.g., CSS `text-transform: uppercase`) signals this to the user. A user typing `brca1` sees their input remain lowercase and may assume it failed for case-related reasons if results are unexpected.

### 3. `searchType` State Leaks Into Guided Search Mode

When the user switches from standard search to Guided Search, `showSearch('guided')` deletes the `searchType` route query param. However, the component's reactive `searchType` data property is not explicitly reset. If the user had previously selected a non-HGVS tab (e.g., "dbSNP rsID"), `this.searchType` retains that value.

When Guided Search completes, `fetchGuidedSearchResults()` builds an HGVS string and delegates to `fetchDefaultSearchResults()`. If `fetchDefaultSearchResults()` reads `this.searchType` rather than forcing `'hgvs'`, the constructed HGVS string is submitted as a dbSNP rsID (or other identifier type), resulting in an empty or nonsensical result set — "breaking in a weird way" with no error message to the user.

---

## Proposed Behavior

1. **Homepage link**: Add a single unobtrusive entry point to Guided Search in the "Explore datasets" or hero quick-links area on `HomeScreen.vue`. A short text link such as "Not sure of the variant identifier? Try guided search →" linking to `/mavemd?mode=guided` is sufficient and avoids cluttering the layout.

2. **Case insensitivity UX**: Apply `text-transform: uppercase` CSS to the gene symbol input field so the user sees their input normalized in real time, matching how it is submitted. No logic change is needed since `.toUpperCase()` is already applied before submission.

3. **Search type reset on guided entry**: When `showSearch('guided')` is called, explicitly reset `this.searchType` to `'hgvs'` (or the default search type constant) in addition to deleting the `searchType` query param. Ensure `fetchGuidedSearchResults()` either passes `'hgvs'` explicitly to `fetchDefaultSearchResults()` or that `this.searchType` is guaranteed to be `'hgvs'` at the time of the guided search submission.

---

## Acceptance Criteria

- [ ] A link or button on `HomeScreen.vue` navigates the user to `/mavemd?mode=guided` (Guided Search mode); it does not significantly add visual weight to the existing hero section.
- [ ] Typing a lowercase gene symbol (e.g., `brca1`) in the Guided Search input displays it in uppercase in the input field, confirming to the user that the search is case-insensitive.
- [ ] Completing a Guided Search after having previously selected any non-HGVS tab in standard search mode returns correct HGVS-based results, identical to the behavior when the user has never changed the search type.
- [ ] No regression in standard search behavior when switching back from Guided to default mode.

---

## Implementation Notes

- **Bug root cause (item 3)**: In `SearchVariantsScreen.vue`, inside `showSearch('guided')`, add `this.searchType = defaultSearchType` (use whatever constant or default value is used elsewhere). Additionally, audit `fetchDefaultSearchResults()` to confirm whether it reads `this.searchType` at call time; if so, either pass the type as an argument or assert that it is always `'hgvs'` when called from the guided path.
- **Case transform (item 2)**: The fix is purely presentational — add `style="text-transform: uppercase"` (or a utility class) to the gene symbol `InputText` in the guided search form. No logic changes are required.
- **Homepage link (item 1)**: The existing quick-links in `HomeScreen.vue` use `RouterLink` components with `/search?...` or `/mavemd?searchType=...` hrefs. A minimal text-link styled consistently with the surrounding "Find a variant with functional data via" copy pointing to `/mavemd?mode=guided` avoids adding a new visual component entirely.

bencap:
This pull request introduces several improvements to the search experience in the variant search screens, focusing on usability and data normalization. The most important changes include normalizing gene symbol input to uppercase, enhancing the guided search flow, and clarifying search options for users.

**Guided Search and User Experience Improvements:**

* The search instructions on the `HomeScreen` were updated to clarify available search types, add a "guided search" option for users without a variant identifier, and improve readability.
* When entering a gene symbol in the guided search, the input is automatically normalized to uppercase, ensuring consistent query parameters and a better user experience. [[1]](diffhunk://#diff-f588bed0e1f93a557eeb6be191d51dc5a16cde2a00d249de4bd32ddac158b585R135) [[2]](diffhunk://#diff-f588bed0e1f93a557eeb6be191d51dc5a16cde2a00d249de4bd32ddac158b585L852-R860)

**Guided Search Logic Enhancements:**

* The guided search now explicitly sets `searchType` to `'hgvs'` when switching modes, ensuring the correct search flow is called after switching between modes.
* The `fetchDefaultSearchResults` method was updated to accept an optional forced search type, allowing more flexible and accurate search queries during guided search. [[1]](diffhunk://#diff-f588bed0e1f93a557eeb6be191d51dc5a16cde2a00d249de4bd32ddac158b585L1030-R1040) [[2]](diffhunk://#diff-f588bed0e1f93a557eeb6be191d51dc5a16cde2a00d249de4bd32ddac158b585L1291-R1300)

### DIFF
--- a/src/components/screens/HomeScreen.vue
+++ b/src/components/screens/HomeScreen.vue
@@ -81,10 +81,12 @@
           <p class="mt-2 text-xs text-text-muted">
             Find a variant with functional data via
             <router-link class="text-link" to="/mavemd?searchType=hgvs">HGVS</router-link>,
-            <router-link class="text-link" to="/mavemd?searchType=clinGenAlleleId">ClinGen CAId</router-link>,
-            <router-link class="text-link" to="/mavemd?searchType=clinVarVariationId">ClinVar ID</router-link>,
-            <router-link class="text-link" to="/mavemd?searchType=dbSnpRsId">dbSNP rsid</router-link>, or
-            <router-link class="text-link" to="/mavemd?searchType=vrsDigest">VRS digest</router-link>.
+            <router-link class="text-link" to="/mavemd?searchType=vrsDigest">VRS digest</router-link>, and more.
+            <br />
+            No variant identifier?
+            <router-link class="text-link" :to="{name: 'mavemd', query: {mode: 'guided'}}"
+              >Try guided search &rarr;</router-link
+            >
           </p>
         </div>
       </section>
--- a/src/components/screens/SearchVariantsScreen.vue
+++ b/src/components/screens/SearchVariantsScreen.vue
@@ -132,6 +132,7 @@
                   v-model="inputGene"
                   class="w-full !rounded-none !border-none !bg-transparent !pt-4 !pb-2 !shadow-none"
                   placeholder="e.g. BRCA1"
+                  style="text-transform: uppercase"
                 />
               </div>
               <div class="guided-select-cell relative flex items-center bg-white">
@@ -849,7 +850,14 @@ export default defineComponent({
         this.inputVariantType = typeof newVal === 'string' ? newVal : ''
       }
     },
-    inputGene() {
+    inputGene(newVal) {
+      const normalized = newVal?.toUpperCase() ?? null
+      if (newVal && newVal !== normalized) {
+        // Setting inputGene re-triggers this watcher with the normalized value, return
+        // early to avoid syncing query params twice.
+        this.inputGene = normalized
+        return
+      }
       this.syncGuidedQueryParams()
     },
     inputVariantType(newVal, oldVal) {
--- a/src/components/screens/SearchVariantsScreen.vue
+++ b/src/components/screens/SearchVariantsScreen.vue
@@ -1000,6 +1000,7 @@ export default defineComponent({
         delete query.search
         delete query.searchType
         query.mode = 'guided'
+        this.searchType = 'hgvs'
       } else {
         delete query.gene
         delete query.variantType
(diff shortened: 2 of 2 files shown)


bencap:
Opens #683.

PR #707, opened 2026-08-06
Davereinhart/gnomad id search

### LINKED ISSUE #704: feat: add gnomAD genomic coordinate search to MaveMD
## Summary

Manuscript review asked for gnomAD-style variant lookup (`X-41334274-A-C`), noting it is increasingly standard for clinical scientists. The ClinGen Allele Registry already exposes this directly as a `gnomAD_4.id` query, so this is a new search type wired to an existing
resolution path and should be trivial to support.

## Problem

MaveMD offers five search types: HGVS, ClinGen Allele ID, dbSNP rsID, ClinVar Variation ID, and VRS digest. A clinical scientist holding a gnomAD-style coordinate has to convert it to HGVS by hand before they can search, even though the registry we already query accepts the
coordinate as-is.

## Proposed behavior

- Add a registry client function alongside the existing ones in the ClinGen allele client, querying the plural `/alleles` endpoint with the `gnomAD_4.id` parameter. The same shape as the existing `dbSNP.rs` and `ClinVar.variationId` lookups. Documented at
  https://reg.clinicalgenome.org/redmine/projects/registry/genboree_registry/alleles?gnomAD_4.id=5-112043362-G-A
- Add a sixth entry to the MaveMD search type options with a validating pattern and a worked example. Accepted value form is `chrom-pos-ref-alt` with no `chr` prefix, e.g. `5-112043362-G-A`, `X-41334274-A-C`.
- Normalize a leading `chr` if present rather than rejecting it. Users will paste it from browsers, and stripping it is cheaper than a support question.
- Teach the global nav search's variant-routing helper to recognize the coordinate form and route to MaveMD with the new search type. The pattern does not collide with the existing ones: HGVS requires a colon, rsID requires the `rs` prefix, ClinGen requires `CA`/`PA`, and bare integers already route to ClinVar.
- Reuse the existing invalid-search toast pattern, with the coordinate example in the detail text.

## Acceptance criteria

- Searching `X-41334274-A-C` via the new MaveMD search type returns the same measurements as searching the equivalent HGVS string for that variant.
- The same query typed into the global nav search routes to MaveMD and returns those results.
- `chr5-112043362-G-A` and `5-112043362-G-A` behave identically.
- A malformed coordinate produces the standard invalid-search toast including a valid example, and issues no registry request.
- A well-formed coordinate with no registry match produces an explicit not-found message rather than a silently empty result list.

davereinhart:
This pull request introduces several improvements and new features to the variant search experience, particularly around searching by gnomAD variant IDs and handling gene symbol searches. It also includes UI enhancements and code refactoring for clarity and maintainability.

### Variant Search Improvements

* Added support for searching by gnomAD variant IDs (`getAlleleByGnomad`), including automatic detection of the correct genome assembly (GRCh38 or GRCh37), fallback logic, and user interface to switch assemblies if needed. The UI now clearly reports which assembly was used and allows users to retry with the other assembly. [[1]](diffhunk://#diff-2bdd422ef0e03bd3a9acc87beb04f6276fb18f0297f424945a6ca626575d1472R39-R83) [[2]](diffhunk://#diff-f588bed0e1f93a557eeb6be191d51dc5a16cde2a00d249de4bd32ddac158b585R823-R826) [[3]](diffhunk://#diff-f588bed0e1f93a557eeb6be191d51dc5a16cde2a00d249de4bd32ddac158b585R289-R341)
* Enhanced the "Any" search option to detect the type of identifier entered and report how the input was interpreted. This provides better feedback to users about their query. [[1]](diffhunk://#diff-f588bed0e1f93a557eeb6be191d51dc5a16cde2a00d249de4bd32ddac158b585R270-R279) [[2]](diffhunk://#diff-f588bed0e1f93a557eeb6be191d51dc5a16cde2a00d249de4bd32ddac158b585R823-R826)

### Gene Symbol Search Improvements

* Improved gene symbol searches: when a valid gene symbol is entered, users are taken directly to the gene page, bypassing the variant search screen, which streamlines navigation and avoids unnecessary history entries.

### API and Data Structure Updates

* Updated the ClinGen API methods: `getAlleleByDbSnp` and `getAlleleByClinVar` now return arrays of alleles instead of a single allele, improving consistency with the API and supporting multiple results.
* Added a comprehensive test suite for the new `getAlleleByGnomad` function, covering various edge cases and error handling.

### UI and Styling Enhancements

* Added new color variables for search types (including gnomAD and gene) for consistent theming across the application.
* Refactored variant result cards to render as plain text when a ClinGen ID is not available, improving clarity for unregistered variants. [[1]](diffhunk://#diff-f588bed0e1f93a557eeb6be191d51dc5a16cde2a00d249de4bd32ddac158b585L298-R368) [[2]](diffhunk://#diff-f588bed0e1f93a557eeb6be191d51dc5a16cde2a00d249de4bd32ddac158b585L325-R395)

### Default Search Behavior

* Changed the default search type on the home screen from "HGVS" to "Any", making the search experience more flexible and user-friendly. [[1]](diffhunk://#diff-4038d17cf80b83a85e6759f9dfbd50c445cad55b83e21c96f28eaf1d8856fec4L262-R263) [[2]](diffhunk://#diff-4038d17cf80b83a85e6759f9dfbd50c445cad55b83e21c96f28eaf1d8856fec4L274-R278)

These changes collectively enhance the search experience, provide clearer feedback to users, and improve code maintainability.

### DIFF
--- a/src/assets/app.css
+++ b/src/assets/app.css
@@ -113,8 +113,8 @@
   --color-badge-alert: #fef3c7;
 
   /* ── Search Types ──────────────────────────────────────────── */
-  --color-any: #5c6b7a;
-  --color-any-light: #eceff2;
+  --color-hgvs: #5c6b7a;
+  --color-hgvs-light: #eceff2;
   --color-dbsnp: #e6b84d;
   --color-dbsnp-light: #fdf3d7;
   --color-clinvar: #5aafa0;
--- a/src/assets/app.css
+++ b/src/assets/app.css
@@ -123,6 +123,8 @@
   --color-ga4gh-light: #e3f0f9;
   --color-gnomad: #5b4ba8;
   --color-gnomad-light: #ece9f7;
+  --color-gene: #b5446e;
+  --color-gene-light: #fbeaf1;
 
   /* ── Measurement Types (variant screen) ────────────────────── */
   --color-nucleotide: #2e7d32;
--- a/src/assets/app.css
+++ b/src/assets/app.css
@@ -119,6 +119,8 @@
   --color-clinvar-light: #c8ece5;
   --color-ga4gh: #0f6ca4;
   --color-ga4gh-light: #e3f0f9;
+  --color-gnomad: #5b4ba8;
+  --color-gnomad-light: #ece9f7;
 
   /* ── Measurement Types (variant screen) ────────────────────── */
   --color-nucleotide: #2e7d32;
--- a/src/assets/app.css
+++ b/src/assets/app.css
@@ -113,6 +113,8 @@
   --color-badge-alert: #fef3c7;
 
   /* ── Search Types ──────────────────────────────────────────── */
+  --color-any: #5c6b7a;
+  --color-any-light: #eceff2;
   --color-dbsnp: #e6b84d;
   --color-dbsnp-light: #fdf3d7;
   --color-clinvar: #5aafa0;
--- a/src/components/screens/SearchVariantsScreen.vue
+++ b/src/components/screens/SearchVariantsScreen.vue
@@ -286,9 +286,12 @@
           <span aria-hidden="true">&larr;</span> New search
         </button>
       </div>
-      <!-- Which reference genome the gnomAD ID was read under -->
+      <!--
+        Which reference genome the gnomAD ID was read under. A reading that hit a mismatch says so in the empty state
+        below instead, which can explain what went wrong rather than implying the coordinates were understood.
+      -->
       <div
-        v-if="gnomadInterpretation"
+        v-if="gnomadInterpretation && !gnomadInterpretation.mismatch"
         class="mb-5 flex flex-col gap-2.5 rounded-lg border border-gray-200 bg-gray-50 px-4 py-3 sm:flex-row sm:items-center sm:justify-between"
       >
         <div class="min-w-0">
@@ -320,11 +323,18 @@
           A resolved reading always yields an allele, registered or not, so an empty result alongside an attempted
           reading means the coordinates don't exist in that assembly rather than that MaveDB lacks the data.
         -->
-        <template v-if="gnomadInterpretation">
+        <template v-if="gnomadInterpretation?.mismatch">
           <div class="text-base font-semibold text-gray-600">Not {{ gnomadInterpretation.name }} coordinates</div>
-          <div class="mt-1.5 text-sm text-gray-500">
-            {{ searchText }} does not match the {{ gnomadInterpretation.name }} reference sequence at that position.
+          <div v-if="gnomadInterpretation.mismatch === 'reference'" class="mt-1.5 text-sm text-gray-500">
+            {{ gnomadInterpretation.id }} does not match the {{ gnomadInterpretation.name }} reference sequence at that
+            position.
           </div>
+          <div v-else class="mt-1.5 text-sm text-gray-500">
+            {{ gnomadInterpretation.id }} lies past the end of that chromosome in {{ gnomadInterpretation.name }}.
+          </div>
+          <code v-if="gnomadInterpretation.hgvs" class="mt-1 block font-mono text-xs text-gray-400">
+            {{ gnomadInterpretation.hgvs }}
+          </code>
           <button
             class="mt-3 cursor-pointer rounded-full border-[1.5px] px-3 py-1 text-xs font-semibold"
             :style="searchColorStyle('gnomadId')"
@@ -820,8 +830,15 @@ export default defineComponent({
       inputAlternateAllele: null as string | null,
       allAlleleOptions: ALLELE_OPTIONS,
       alleles: [] as AlleleResult[],
-      /** Which assembly the current gnomAD ID was read under; null when the search wasn't a gnomAD ID search. */
-      gnomadAssembly: null as GenomeAssembly | null,
+      /**
+       * The gnomAD reading currently on screen: the ID exactly as searched, the assembly it was read (or attempted)
+       * under, and how the registry answered. A null mismatch means no verdict on the assembly — either the reading
+       * resolved, or it failed for a reason that says nothing about the coordinates.
+       *
+       * This records what the last search did, rather than tracking the form: the results must not restate
+       * themselves as the user types a new identifier into the box.
+       */
+      gnomadReading: null as {id: string; assembly: GenomeAssembly; mismatch: 'reference' | 'position' | null} | null,
       /** What an "Any" search resolved the string to; null when the search type was chosen explicitly. */
       detectedSearchType: null as string | null,
       nucleotideScoreSetListIsExpanded: [] as Array<boolean>,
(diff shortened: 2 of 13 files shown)


src/assets/app.css
```diff
@@ -113,12 +113,18 @@
   --color-badge-alert: #fef3c7;
 
   /* ── Search Types ──────────────────────────────────────────── */
+  --color-any: #5c6b7a;
+  --color-any-light: #eceff2;
```
bencap: My first thought looking at the changes was that the default on our landing page should probably be the sage color. I don't think anything is holding us to HGVS being sage, so just to keep the initial load centered on the brand color I would switch the any and hgvs colors.

src/components/screens/SearchVariantsScreen.vue
```diff
@@ -985,11 +1084,19 @@ export default defineComponent({
...
+      const geneSymbol = geneSymbolSearchTarget(example, searchType)
+      if (geneSymbol && geneSymbolRegex.test(geneSymbol)) {
+        this.router.push({name: 'gene', params: {symbol: geneSymbol}})
+        return
+      }
       this.searchType = searchType
       this.showSearch('default')
       this.searchText = example
-      this.defaultSearch()
+      await this.defaultSearch()
```
bencap, line 1126:
```suggestion
      await this.$nextTick()
      await this.defaultSearch()
```

Selecting an example from a different search type to the one selected searches correctly, but doesn't show the text in the box. Awaiting the next tick allows watchers on the search type to fire immediately so they don't flush the box clean after the below search.

src/components/screens/SearchVariantsScreen.vue
```diff
@@ -1024,9 +1133,35 @@ export default defineComponent({
       this.router.replace({query})
       this.clearSearch()
     },
-    defaultSearch: async function () {
+    defaultSearch: async function (forcedAssembly?: GenomeAssembly) {
       if (this.searchText) this.searchText = this.searchText.trim()
+
+      // A gene symbol navigates away instead of producing results here, so it is handled before this screen records
+      // the search in its query params. The search is also stripped from the entry being left behind: were it to
+      // remain, returning here would re-run it on mount and bounce the user straight back to the gene page.
```
bencap, line 1168: Nice!

src/components/screens/SearchVariantsScreen.vue
```diff
@@ -748,6 +820,10 @@ export default defineComponent({
       inputAlternateAllele: null as string | null,
       allAlleleOptions: ALLELE_OPTIONS,
       alleles: [] as AlleleResult[],
+      /** Which assembly the current gnomAD ID was read under; null when the search wasn't a gnomAD ID search. */
+      gnomadAssembly: null as GenomeAssembly | null,
```
bencap: It seems to me like the template and the computed actually need slightly more than just the GenomeAssembly here.

Searching with `3-198100000-A-C` can trace the whole flow. Enter the search term, click search, and we get a reasonable no variants found response for GRCh38. GRCh37 chr3 ends ~78 kb short of that position. Clicking `Read as GRCh37` gets a `IncorrectHgvsPosition` error, the error handler at line 1278 rethrows the error, and the generic toast fires with a message that tells us what went wrong. However the template renders `Not GRCh37 coordinates — 3-198100000-A-C does not match the GRCh37 reference sequence at that position.` even though the search couldn't have consulted a reference there. A similar symptom of this is that typing more text into the search box will have it render live in the template.

Ultimately, the template wants to make a claim about what happened the last time the user hit enter, not a claim about current state of the model. The comment on lines 1161-62 implies this, but doesn't give the template the full power it needs to reconstruct the attempted search and tell the user what happened.

My suggestion would be to retype `gnomadAssembly` as something like:
```typescript
/**
 * The gnomAD reading currently on screen: the ID exactly as searched, the assembly it was read (or attempted)
 * under, and how the registry answered. `mismatch: null` means we never got a verdict — the reading resolved,
 * or the request failed for a reason unrelated to the assembly.
 */
gnomadReading: null as {
  id: string
  assembly: GenomeAssembly
  mismatch: 'reference' | 'position' | null
} | null,
```

There would then be two write paths:
```typescript
// line 1163 — forced
this.gnomadReading = forcedAssembly && this.searchText
  ? {id: this.searchText, assembly: forcedAssembly, mismatch: null}
  : null

...

// line 1268 — resolved
this.gnomadReading = {id: searchStr, assembly: gnomadResult.assembly, mismatch: null}
```

Then, the computed can pass these through to the template:
```typescript
gnomadInterpretation() {
  if (!this.gnomadReading) return null
  const {id, assembly, mismatch} = this.gnomadReading
  return {
    id,                                                        // for line 326, rather than using the current model search str
    name:      GENOME_ASSEMBLY_NAMES[assembly],
    hgvs:      gnomadIdToHgvs(id, assembly),
    otherName: GENOME_ASSEMBLY_NAMES[otherAssembly(assembly)],
    mismatch                                                   // for the gate + copy
  }
}
```

Then we can leverage the error handling already written on lines 1274-1303 to tell the template what happened, instead of having that information limited to the toast:
```typescript
const {data} = getErrorResponse(error)

// understand the mismatch type
const mismatch =
  data?.errorType === 'IncorrectReferenceAllele' ? 'reference' :
  data?.errorType === 'IncorrectHgvsPosition'    ? 'position'  : null

// the error is unrelated to the assembly
if (!mismatch) throw error

// thread the mismatch type to the template
if (this.gnomadReading) this.gnomadReading.mismatch = mismatch
```

Now the template has full context on what was searched, what it was searched against, and why it failed without having to reach into the model. We can use `gnomadInterpretation.id` rather than `searchText` so that typing in the search box doesn't pollute the template. But more usefully, we can better tailor the template error message like we are doing already with the toast:
```html
<!-- line 318 -->
<template v-if="gnomadInterpretation?.mismatch">
  <div>Not {{ gnomadInterpretation.name }} coordinates</div>
  <div v-if="gnomadInterpretation.mismatch === 'reference'">
    {{ gnomadInterpretation.id }} does not match the {{ gnomadInterpretation.name }} reference sequence at that position.
  </div>
  <div v-else-if="gnomadInterpretation.mismatch === 'reference'">
    {{ gnomadInterpretation.id }} lies past the end of that chromosome in {{ gnomadInterpretation.name }}.
  </div>
  <div v-else>
    uh oh...
  </div>
</template>
```

That ended up being a longer comment than intended, so hopefully it all makes sense. Happy to chat about it if not though.

bencap (commented):
Thanks Dave! We should address the second and fourth comments below prior to merging, but it's going to be really nice to have a search field that handles all our identifiers from a single catch-all search.

I didn't really have a great idea for the pills spilling over to the second line. In the long run, I think we'll probably be happiest if we unify the homepage and the MaveMD page into a shared MvVariantSearchBar component with the dropdown style selector. Don't feel like you have to make that change here though, its fine with me if we leave the pills on multiple lines and track that as a follow up issue.

I was also thinking about the regEx you built out here and it reminded me of the `routeToVariantSearchIfVariantIsSearchable` function in `search.ts`. I thought it might make sense to extract or import `detectSearchType` there and use it within the routing function. The only thing that gives me pause is thinking that plain integers should remain on that page and that gene symbols probably shouldn't preemptively route to the gene page within the data set search flow. This also seems fine to defer if it's a pain.

davereinhart:
Thanks for the review and detailed comments @bencap! I addressed each of them and also added 0085ad38f75ce1c43047873393d91f1d15f66f24 which was not covered here. It is a global change, so wanted to check if you know of any cases where we would not want the vertical scroll position reset when navigating to a new page.

I can look into extracting `detectSearchType` too, but the cases you mentioned do sound like it might need more changes before using it in the routing function so it might be worth treating as a separate issue.

bencap (approved):
Looks good, thanks Dave! I can open a couple issues for the search RegEx harmonization and consolidating the components.

PR #709, opened 2026-08-10
Debug the dropdown menu of superseding score set shows private score sets problem.

### LINKED ISSUE #708: The dropdown menu of superseding score set has private score sets in creating a new score set

EstelleDa:


src/components/screens/ScoreSetCreator.vue
```diff
@@ -473,7 +473,11 @@ export default {
     const publications = usePublicationIdentifiers()
     const licenseSearch = useAutocomplete<License>('/licenses/active')
 
-    const extractScoreSets = (data: unknown) => ((data as Record<string, unknown>)?.scoreSets as ShortScoreSet[]) || []
+    const extractScoreSets = (data: unknown): ShortScoreSet[] => {
```
bencap: Perfect, looks good. `supersededSearch` and `metaAnalyzesSearch` both want public score sets only. My only suggestion is that we rename `extractScoreSets` to something that makes it clear that we are keeping only public score sets now.

bencap (approved):

PR #716, opened 2026-08-13
Feature/estelle/721/superseding calibration

EstelleDa:
Front end of https://github.com/VariantEffect/mavedb-api/issues/721

### DIFF
--- a/src/components/calibration/CalibrationEditor.vue
+++ b/src/components/calibration/CalibrationEditor.vue
@@ -37,45 +37,21 @@
     @toggle-boundary="onToggleBoundary"
     @toggle-infinity="onToggleInfinity"
     @toggle-oddspaths="onToggleOddspaths"
-    @update:baseline-score="
-      draft.baselineScore = $event
-      markChanged()
-    "
-    @update:baseline-score-description="
-      draft.baselineScoreDescription = $event
-      markChanged()
-    "
+    @update:baseline-score="draft.baselineScore = $event; markChanged()"
+    @update:baseline-score-description="draft.baselineScoreDescription = $event; markChanged()"
     @update:class-based="classBased = $event"
     @update:classification-field="onClassificationFieldUpdate"
-    @update:evidence-sources="
-      draft.evidenceSources = $event
-      markChanged()
-    "
+    @update:evidence-sources="draft.evidenceSources = $event; markChanged()"
     @update:evidence-strength="onEvidenceStrengthUpdate"
     @update:is-superseding-calibration="isSupersedingScoreCalibration = $event"
-    @update:method-sources="
-      draft.methodSources = $event
-      markChanged()
-    "
-    @update:notes="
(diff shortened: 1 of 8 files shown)


src/components/calibration/CalibrationEditor.vue
```diff
@@ -38,11 +43,13 @@
     @update:classification-field="onClassificationFieldUpdate"
     @update:evidence-sources="draft.evidenceSources = $event; markChanged()"
     @update:evidence-strength="onEvidenceStrengthUpdate"
+    @update:is-superseding-calibration="isSupersedingScoreCalibration = $event"
```
bencap, line 46: This handler and the one on line 52 don't call `markChanged()`, where every other field handler in this template does. `markChanged` is what recomputes the editor's dirty state, so toggling supersession or picking a calibration is a change the editor doesn't know happened — the unsaved-changes prompt won't fire on navigation, and if the save control is gated on dirty state you'd have to touch an unrelated field to enable it.

src/components/calibration/CalibrationFields.vue
```diff
@@ -32,6 +32,64 @@
         </div>
       </div>
 
+      <!-- Superseding toggle -->
```
bencap, line 35: This is the authoring half of #721 and it reads well. The display half isn't here though — the deprecation notice on a superseded calibration, the "supersedes" notice on the superseding one, and being able to follow a chain without copying URNs by hand. Is that planned as a follow-up, or was it meant to be in this PR?

Asking partly because it may not be buildable yet. The API nulls `supersededCalibration` and `supersedingCalibration` on every read path — see https://github.com/VariantEffect/mavedb-api/pull/769#discussion_r3834204942 — so there's nothing arriving for those components to render.

src/schema/openapi.d.ts
```diff
@@ -4571,6 +4573,8 @@ export interface components {
       methodSources: components["schemas"]["PublicationIdentifierCreate"][];
       /** Calibrationmetadata */
       calibrationMetadata?: Record<string, never> | null;
+      supersededCalibration?: components["schemas"]["ShorterScoreCalibration"] | null;
```
bencap, line 4576: I think this file was edited by hand rather than regenerated, and it's drifted from the API in three ways:

- `/score-calibrations/me/search` has no path entry at all, which is why `CalibrationEditor` calls it as a raw string and casts the response through `Record<string, unknown>`
- `supersededCalibrationUrn` — the field the create endpoint actually accepts, and the one the draft sends — doesn't appear anywhere in the file
- this hunk adds `supersededCalibration` and `supersedingCalibration` to `ScoreCalibrationCreate`, but those are response-only fields on `SavedScoreCalibration`; the create model has no such fields

Worth regenerating against the running API with `npx openapi-typescript http://localhost:8002/openapi.json -o src/schema/openapi.d.ts`

